PR #7, opened 2024-08-13
ISD-2183 Add certificates integration tests

Thanhphan1147:
Applicable spec: ISD163 (internal)

Add unit and integration tests for the new `tls_relation` module

### Checklist


tests/integration/test_action.py
```diff
@@ -0,0 +1,32 @@
...
+from .conftest import TEST_EXTERNAL_HOSTNAME_CONFIG
+
+
+@pytest.mark.abort_on_fail
+async def test_get_certificate_action(
+    configured_application_with_tls: Application,
+):
+    """Deploy the charm with valid config and tls integration.
+
+    Assert on valid output of get-certificate.
```
jdkandersson: Perhaps one of our linters is missing? These tests should be described in the arrange/act/assert format which is checked for using https://github.com/jdkandersson/flake8-test-docs

Thanhphan1147: I just double-checked and we do have `flake8-test-docs` installed in the `lint` step : https://github.com/canonical/haproxy-operator/blob/6445921388de49dee8fc1a29fda0d4e9dc0b7672/tox.ini#L44

Thanhphan1147: I'm still not sure why it's not detected in the lint step but I'll update the docstrings to the arrage/act/assert format

Thanhphan1147: I updated the docstring of integration tests to follow the format. I haven't looked into why the plugins is not working properly yet, but I'll find the time to look into it.

tests/integration/test_action.py
```diff
@@ -0,0 +1,34 @@
...
+    arrange: Deploy the charm with valid config and tls integration.
+    act: Run the get-certificate action and run a sh command to check
+    the cert location on the unit.
+    assert: The output of both operations are valid.
+    """
+    action = await configured_application_with_tls.units[0].run_action(
+        "get-certificate", hostname=TEST_EXTERNAL_HOSTNAME_CONFIG
+    )
+    await action.wait()
+    assert action.results
```
amandahla: What about checking the results content and also adding failure scenarios for example if you run the action without the relation?

tests/integration/test_charm.py
```diff
@@ -2,11 +2,34 @@
...
 async def test_deploy(application: Application):
-    """Deploy the charm."""
-    assert application
+    """
+    arrange: Deploy the charm.
+    act: Send a GET request to the unit's ip address.
+    assert: The charm correctly response with the default page.
+    """
+    status: FullStatus = await application.model.get_status([application.name])
+    unit_status: UnitStatus = next(iter(status.applications[application.name].units.values()))
```
amandahla: I'll send you a MM thread link to more options for getting unit IP address in integration tests :)

Thanhphan1147: updated the code to use that method instead.

tests/integration/test_action.py
Thanhphan1147: Thx, I added a failure case and also update the test to test the correct output.

tests/integration/test_action.py
```diff
@@ -0,0 +1,38 @@
...
+    the cert location on the unit.
+    assert: The output of both operations are valid.
+    """
+    action = await configured_application_with_tls.units[0].run_action(
+        "get-certificate", hostname=TEST_EXTERNAL_HOSTNAME_CONFIG
+    )
+    await action.wait()
+    assert "-----BEGIN CERTIFICATE-----" in action.results.get("certificate")
+
+    with pytest.raises(Exception):
```
amandahla: I didnt follow this one...Why rerunning the action but this time with pytest.raises(Exception)?

Thanhphan1147: Sorry I wasn't being clear here. Maybe I needed to add a comment explaining it. Basically the second time we run the action without the `hostname` parameter, and if the event fails pytest should raise an `Exception`

alithethird: Maybe the second running can be done by parameterizing the test? This way it would be more clear which test fails.

Thanhphan1147: Since the logic differs quite a bit, I moved the error case to a second test

amandahla (approved):
LGTM

tests/integration/test_config.py
```diff
@@ -6,8 +6,11 @@
 
 
 async def test_config(application: Application):
-    """Deploy the charm, set global-maxconn config and
-    verify that the correct value is rendered.
+    """
+    arrange: Deploy the charm.
+    act: Update the charm config to a valid value and then an invalid value.
```
alithethird, line 11: First an invalid value then a valid value.

tests/unit/conftest.py
```diff
@@ -25,3 +29,91 @@ def systemd_mock_fixture(monkeypatch: pytest.MonkeyPatch):
...
+    certificates_relation_data: dict[str, str],
+) -> Harness:
+    """Mock certificates integration."""
+    harness.set_leader()
+    harness.update_config({"external-hostname": TEST_EXTERNAL_HOSTNAME_CONFIG})
+    relation_id = harness.add_relation(
+        "certificates", "self-signed-certificates", app_data=certificates_relation_data
+    )
+    harness.update_relation_data(
+        relation_id, harness.model.app.name, {f"csr-{TEST_EXTERNAL_HOSTNAME_CONFIG}": "csr"}
```
alithethird, line 117: Why not create the relation data at once? Why create and update?

alithethird (commented):
Left 2 comments also you have a typo in `src/haproxy.py` `s/haporxy/haproxy`.
I have a couple more notes:
1- I think you can put the `TLSInformation.validate()`  in a decorator. I think you can even put it in the `validate_config_and_tls` decorator and activate with a parameter like `validate_info`. You are calling it in 4 different functions.
2- Why use `typing.cast(str, ...)` instead of `str(...)`? As far as I know `typing.cast()` doesn't actually cast it, it just silences the mypy etc. [doc](https://docs.python.org/3/library/typing.html#typing.cast)

tests/unit/conftest.py
Thanhphan1147: I'm updating the relation data bag for the haproxy charm at this point to add the csr, and I believe you can only update the relation data for the remote app during creation.

Thanhphan1147: Just checked to see if pytest ops has been updated to allow you to add the current app relation data during creation but it seems like it's still not possible

tests/integration/test_config.py
Thanhphan1147: Thanks, I'll update this and the other typos in the next PR

Thanhphan1147:
> Left 2 comments also you have a typo in `src/haproxy.py` `s/haporxy/haproxy`. I have a couple more notes: 1- I think you can put the `TLSInformation.validate()` in a decorator. I think you can even put it in the `validate_config_and_tls` decorator and activate with a parameter like `validate_info`. You are calling it in 4 different functions. 2- Why use `typing.cast(str, ...)` instead of `str(...)`? As far as I know `typing.cast()` doesn't actually cast it, it just silences the mypy etc. [doc](https://docs.python.org/3/library/typing.html#typing.cast)

I didn't put `TLSInformation.validate()` in a decorator because the hook was following the general flow of `initializing the chamstate component for the hook` -> `do business logic` and any errors will be caught and handled by the decorator. In cases where the return value of the charmstate component is not needed, then we'll use the `validate()` method of the charmstate component class. You can imagine different hooks can require any combination of charmstate components so putting all of them in a decorator and add a bunch of parameters is a bit messy. Also it doesn't make much sense if `tls_information = TLSInformation.from_charm(self, self.certificates)` and `TLSInformation.validate(self)` are in different places when they are functionally the same.

For `typing.cast` the main purpose for it is actually to tell mypy that at that point either the cert or the provider cert must exists

alithethird (approved):
LGTM :rocket:

PR #14, opened 2024-09-02
ISD-2184 Implement reverseproxy relation

Thanhphan1147:
Implement support for the `reverseproxy:http` relation. As we prioritize having feature parity with the legacy haproxy charm as well as for the new charm to be a drop-in replacement, part of the code of the legacy charm is added and isolated in the `legacy.py` module (all lint checks disabled).

Supported workflow:
```
juju deploy pollen
juju deploy ./haproxy_amd64.charm
juju relate pollen haproxy
curl <haproxy_unit_ip>
```

### Checklist


Thanhphan1147:
Hi @gregory-schiano, can you take a look at this PR and leave your review? Thanks!

src/haproxy.py
```diff
@@ -79,19 +83,27 @@ def is_active(self) -> bool:
...
-            config: charm config
+            config: The charm config.
+            http_provider: The http interface provider.
         """
         with open("templates/haproxy.cfg.j2", "r", encoding="utf-8") as file:
-            template = Template(file.read(), keep_trailing_newline=True)
-        rendered = template.render(config_global_max_connection=config.global_max_connection)
+            template = Template(
+                file.read(), keep_trailing_newline=True, trim_blocks=True, lstrip_blocks=True
+            )
```
jdkandersson, line 94: Would https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text work here? Pathlib is a more modern file system interface in Python

src/http_interface.py
```diff
@@ -0,0 +1,162 @@
...
+    Attrs:
+        data_provided: Custom event when integration data is provided.
+        data_removed: Custom event when integration data is removed.
+    """
+
+    data_provided = EventSource(HTTPDataProvidedEvent)
+    data_removed = EventSource(HTTPDataRemovedEvent)
+
+
+class _IntegrationInterfaceBaseClass(Object):
```
jdkandersson, line 41: This seems like an abstract base class, is that needed in this case? It looks like this is only used as the base class for one other class?

src/http_interface.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class HTTPProvider(_IntegrationInterfaceBaseClass):
+    """HTTP interface provider class to be instantiated by the haproxy-operator charm.
+
+    Attrs:
+        on: Custom events that are used to notify the charm using the provider.
+    """
+
+    on = HTTPProviderEvents()  # type: ignore
```
jdkandersson: All `type: ignore` and other linting disable rules require an explanatory comment please

jdkandersson: This is so that future readers know the reason for the disable.

src/legacy.py
jdkandersson, line 1: This code doesn't look great, any plans on what we do about it please?

tests/integration/conftest.py
```diff
@@ -29,7 +31,7 @@ async def model_fixture(ops_test: OpsTest) -> Model:
 @pytest_asyncio.fixture(scope="module", name="charm")
 async def charm_fixture(pytestconfig: pytest.Config) -> str:
     """Get value from parameter charm-file."""
-    charm = pytestconfig.getoption("--charm-file")
+    charm = typing.cast(str, pytestconfig.getoption("--charm-file"))
```
jdkandersson: What is the reason for needing this cast please?

src/haproxy.py
```diff
@@ -79,19 +83,27 @@ def is_active(self) -> bool:
...
+
+        services_dict = http_provider.get_services_definition()
```
gregory-schiano: Pass the services_dict to the method instead of the http_provider

src/http_interface.py
```diff
@@ -0,0 +1,162 @@
...
+logger = logging.getLogger()
+SERVICES_CONFIGURATION_KEY = "services"
+
+
+class HTTPDataProvidedEvent(RelationEvent):
+    """Event representing that http data has been provided."""
+
+
+class HTTPDataRemovedEvent(RelationEvent):
+    """Event representing that http data has been removed."""
```
gregory-schiano: I would call them `HTTPBackendAvailable` and `HTTPBackendDeparted` or `HTTPBackendRemoved`

src/http_interface.py
```diff
@@ -0,0 +1,162 @@
...
+        observe(charm.on[relation_name].relation_broken, self._on_relation_broken)
+
+    def _on_relation_joined(self, _: RelationJoinedEvent) -> None:
+        """Abstract method to handle relation-joined event."""
+
+    def _on_relation_changed(self, _: RelationChangedEvent) -> None:
+        """Abstract method to handle relation-changed event."""
+
+    def _on_relation_broken(self, _: RelationBrokenEvent) -> None:
+        """Abstract method to handle relation-changed event."""
```
gregory-schiano: Add `@abstractmethod` decorator and raise `NotImplementedError`

src/http_interface.py
```diff
@@ -0,0 +1,162 @@
...
+
+    @property
+    def relations(self) -> list[Relation]:
+        """The list of Relation instances associated with the charm."""
+        return list(self.charm.model.relations[self.relation_name])
```
gregory-schiano: Might require being defensive to avoid `KeyError`

src/http_interface.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def _on_relation_joined(self, event: RelationJoinedEvent) -> None:
+        """Handle relation-changed event.
+
+        Args:
+            event: relation-changed event.
+        """
+        event.relation.data[self.charm.unit].update(
+            {
+                "public-address": self.charm.bind_address,  # type: ignore
```
gregory-schiano: `bind_address` in only used in this library, so I'd move the property implementation to the Abstract class

src/http_interface.py
```diff
@@ -0,0 +1,162 @@
...
+            }
+        )
+
+    def _on_relation_changed(self, event: RelationChangedEvent) -> None:
```
gregory-schiano, line 131: To be evaluated:
- Add a property in the class that contains the service dict
- Parse the relation data in this hook and store it in the class attribute
- Manage potential edges cases here

src/legacy.py
```diff
@@ -0,0 +1,441 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+# We ignore all lint errors in the legacy charm code as we've decided
+# to reuse them for the support of the legacy relations
+# flake8: noqa
+# pylint: skip-file
+# mypy: ignore-errors
+# fmt: off
+"""Legacy haproxy module."""
```
gregory-schiano: We should be more explicit on the fact that we took the legacy implementation, and explain the reasoning.
Also eventually comment lines that we had to adapt

src/state/config.py
gregory-schiano, line 1: Seems to be used in legacy to get the unit name, I'd rather use a constant here since we don't really care having each unit getting a different configuration, they're independent

src/haproxy.py
Thanhphan1147: In a PR further down the line I switch to using loaders to be able to manage inheritance, so since this code will be replaced soon, I'll resolved this comment if that's ok.

src/http_interface.py
Thanhphan1147: I have the HTTPRequirer class also planned for the `website` relation, but that has not been implemented yet.

src/legacy.py
Thanhphan1147: I'll add more details to the header, but in short since we want this charm to work as a drop-in replacement to the legacy charm, we need to keep the legacy charm's code isolated here to ensure that the behavior stays the same between the 2 charms (with regards to the reverseproxy and website relation)

src/haproxy.py
Thanhphan1147: There's a later PR that did this but I moved the changes to this PR as well.

src/http_interface.py
Thanhphan1147: thanks, added.

src/http_interface.py
Thanhphan1147: Updated.

src/http_interface.py
Thanhphan1147: Added parsing the relation data in the init method and store it as the `services` instance attribute

src/state/config.py
Thanhphan1147: updated to use a constant and refactored all related methods

src/legacy.py
Thanhphan1147: I added a detailed explanation for the `legacy.py` module

src/http_interface.py
Thanhphan1147: I updated this to using `.get()`

src/http_interface.py
Thanhphan1147: I updated the naming.

src/legacy.py
Thanhphan1147: I added more details to the module

src/http_interface.py
Thanhphan1147: Actually I just removed this as it's not useful

tests/integration/conftest.py
Thanhphan1147: I can't remember why I needed to add this, so I'll remove it for now.

src/charm.py
```diff
@@ -21,7 +21,7 @@
 from ops.charm import ActionEvent, RelationJoinedEvent
 
 from haproxy import HAProxyService
-from http_interface import HTTPDataProvidedEvent, HTTPDataRemovedEvent, HTTPProvider
+from http_interface import HTTPBackendAvailable, HTTPBackendRemoved, HTTPProvider
```
gregory-schiano: Missing Event suffix :)

src/http_interface.py
```diff
@@ -63,29 +65,72 @@ def __init__(self, charm: CharmBase, relation_name: str):
...
+        """
+        raise NotImplementedError
 
     @property
     def relations(self) -> list[Relation]:
         """The list of Relation instances associated with the charm."""
-        return list(self.charm.model.relations[self.relation_name])
+        relations = self.charm.model.relations.get(self.relation_name)
+        if relations is None:
+            return []
```
gregory-schiano: `.get(self.relation_name, [])`

src/http_interface.py
```diff
@@ -63,29 +65,72 @@ def __init__(self, charm: CharmBase, relation_name: str):
...
 
-    on = HTTPProviderEvents()  # type: ignore
+    on = HTTPProviderEvents()
+
+    def __init__(self, charm: CharmBase, relation_name: str):
+        """Initialize the HTTPProvider class and parse the relation data.
+
+        Args:
+            charm: The charm instance
+            relation_name: _description_
```
gregory-schiano: Missing description

Thanhphan1147: I added the description

src/http_interface.py
Thanhphan1147: Thanks, I updated the code to reflect this

src/charm.py
Thanhphan1147: Added the missing Event suffix

pyproject.toml
```diff
@@ -32,6 +35,7 @@ docstring-convention = "google"
 [tool.isort]
 line_length = 99
 profile = "black"
+skip = ["src/legacy.py"]
```
yhaliaw, line 38: Does import sorting need to be disable for `legacy.py`?

Thanhphan1147: I don't think so but my thought process was if all lints are disabled on this module then controlling imports sorting wouldn't make much of a difference, so I just disabled everything including import sorting checks

src/http_interface.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def __init__(self, charm: CharmBase, relation_name: str):
+        """Initialize the HTTPProvider class and parse the relation data.
+
+        Args:
+            charm: The charm instance
+            relation_name: The name of the relation
+        """
+        super().__init__(charm, relation_name)
+        self.services = legacy.generate_service_config(self.get_services_definition())
```
gregory-schiano: There are 2 legacy methods called from this line, which a lot of business logic in them and no error handling in them. I'm wondering if that's a good idea to call this from the constructor

Thanhphan1147: Updated to move this code to a method and remove the init method to let the exception flow to the charm as we discussed.

src/http_interface.py
```diff
@@ -0,0 +1,202 @@
...
+
+        observe = self.framework.observe
+        self.charm: CharmBase = charm
+        self.relation_name = relation_name
+
+        observe(charm.on[relation_name].relation_created, self._on_relation_changed)
+        observe(charm.on[relation_name].relation_joined, self._on_relation_joined)
+        observe(charm.on[relation_name].relation_changed, self._on_relation_changed)
+        observe(charm.on[relation_name].relation_departed, self._on_relation_changed)
+        observe(charm.on[relation_name].relation_broken, self._on_relation_broken)
```
cbartz: It is not strictly conform to https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619 (see section `Event handler names`, each event should have its own handler).

src/http_interface.py
```diff
@@ -0,0 +1,202 @@
...
+        Raises:
+            NotImplementedError: if the abstract method is not implemented.
+        """
+        raise NotImplementedError
+
+    @property
+    def relations(self) -> list[Relation]:
+        """The list of Relation instances associated with the charm."""
+        relations = self.charm.model.relations.get(self.relation_name, [])
+        return relations
```
cbartz: Any reason for having a variable `relations` here?

cbartz (commented):
LGTM, minor comments. I am missing unit tests for the changes. @Thanhphan1147

src/http_interface.py
Thanhphan1147: No particular reason, it was brought over from the traefik lib. I updated the code to return directly

src/http_interface.py
Thanhphan1147: I just removed the other events handlers as they are not very useful anyways.

Thanhphan1147:
@cbartz can I ask you to take a look at the PR again and leave your reviews? I'll add the unit tests in a follow-up PR.

Thanks!

cbartz (approved):

gregory-schiano (approved):

PR #16, opened 2024-09-10
Add ingress integration

Thanhphan1147:
Applicable spec: ISD-163 (internal)

### Overview

When under ingress mode, TLS is forced with a redirect option and backend selection is performed using [ACLs](https://docs.haproxy.org/2.8/configuration.html#7.1) and the use_backend directive. The default routing logic is path-based with support for subdomain-based routing planned to be added.

frontend frontend
    mode http
    bind :80
    bind :443 ssl crt <path-to-cert> (see [TLS termination](https://docs.google.com/document/d/1bfqBGD8Oj24nef389_54NqMf887CDGfyF2hNJ_6fiNY/edit#heading=h.zeq77oqix8mp))
   
    # Redirect HTTP to HTTPS
    http-request redirect scheme https unless { ssl_fc }
    
    acl path_<application_name> path -i -m beg /<model_name>-<application_name>
    use_backend backend-<application_name> if path_<application_name> 
    default_backend default

Where <application_name> and <model_name> are values present in the ingress requirer charm’s application databag. See [charm-relation-interfaces](https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/ingress/v2) for more information.
Backend

Initially the haproxy charm will not support customizing service_options and server_options under ingress mode and any relevant options will be hard-coded. However, making certain service_options and server_options configurable is possible should the need arise.

backend <application_name>
    balance roundrobin
    server <requirer_unit_0> <host>:<port> check
    server <requirer_unit_0> <host>:<port> check
    …

<host> and <port> are data passed by the ingress requirer charm in each of its units. See [charm-relation-interfaces](https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/ingress/v2#requirer-1) for more information.

### Checklist


src/charm.py
```diff
@@ -177,6 +225,57 @@ def _reconcile_certificates(self) -> None:
...
+        self._ingress_provider.publish_url(
+            event.relation, f"https://{tls_information.external_hostname}/{path_prefix}/"
+        )
+
+    @validate_config_and_tls(defer=False, block_on_tls_not_ready=True)
+    def _on_ingress_data_removed(self, _: IngressPerAppDataRemovedEvent) -> None:
+        """Handle the data-removed event."""
+        self._reconcile()
+
+    def _validate_state(self) -> tuple[bool, ProxyMode]:
```
gregory-schiano: Why returning a tuple (very go like) instead of throwing exceptions ?

src/charm.py
```diff
@@ -152,10 +174,36 @@ def _on_get_certificate_action(self, event: ActionEvent) -> None:
...
         """Render the haproxy config and restart the service."""
+        is_state_valid, proxy_mode = self._validate_state()
+        if not is_state_valid:
+            # We don't raise any exception/set status here as it should already be handled
+            # by the _validate_state method
+            return
+
         config = CharmConfig.from_charm(self)
-        self.haproxy_service.reconcile(config)
+        kwargs = {}
```
gregory-schiano: Not a very good variable name, `reconcile_extra_args` ? Or at least `reconcile_kwargs`

src/charm.py
```diff
@@ -152,10 +174,36 @@ def _on_get_certificate_action(self, event: ActionEvent) -> None:
...
+                self._ingress_provider
+            )
+            tls_information = TLSInformation.from_charm(self, self.certificates)
+            kwargs["ingress_requirers_information"] = ingress_requirers_information
+            kwargs["haproxy_crt_dir"] = HAPROXY_CERTS_DIR
+            kwargs["config_external_hostname"] = tls_information.external_hostname
+        if proxy_mode == ProxyMode.LEGACY:
+            kwargs["services"] = self.http_provider.services
+
+        self.haproxy_service.reconcile(proxy_mode, config, **kwargs)
```
gregory-schiano: I'm a bit hesitant here, having a single `reconcile` method that supports both modes could be interesting from an abstraction aspect
But since the attributes are specific per proxy_mode, the haproxy still needs to know the mode, therefore the added value of a single method is very limited
I'd then rather have 2 reconcile method `reconcile_legacy` and `reconcile_ingress` that could call themselves a private `_reconcile` method for the code they have in common

This will avoid the generic `kwargs` dict, and won't need to pass the `proxy_mode`

src/charm.py
gregory-schiano, line 1: The hooks should all have the decorator

src/charm.py
gregory-schiano: The cases where you return `False` are only linked to `ProxyMode.INVALID` therefore just relying on that info will work

src/charm.py
Thanhphan1147: added the decorator to all of the relevant hooks

src/charm.py
Thanhphan1147: removed this entirely

src/charm.py
Thanhphan1147: makes sense, I split the `reconcile` method of the `haproxy_service` module into 3 methods: `reconcile_ingress`, `reconcile_legacy` and `reconcile_default`. They call the shared `render_haproxy_config` method with the appropriate template name and context.

src/charm.py
Thanhphan1147: Thanks, I updated the code to only return the proxymode, and based on the proxymode I called the corresponding `reconcile` method

templates/haproxy_ingress.cfg.j2
```diff
@@ -0,0 +1,26 @@
...
+{% endfor %}
+
+{% for backend in ingress_requirers_information.backends %}
+backend {{ backend.backend_name }}
+    balance roundrobin
+{% for server in backend.servers %}
+    server {{ server.server_name }} {{ server.hostname_or_ip }}:{{ server.port }} check
+{% endfor %}
+{% endfor %}
+{% endblock %}
```
yanksyoon, line 26: can we add a newline here for this config?

templates/haproxy_legacy.cfg.j2
```diff
@@ -0,0 +1,6 @@
+{% extends 'haproxy.cfg.j2' %}
+{% block proxy_configuration %}
+{% for service in services %}
+{{ service }}
+{% endfor %}
+{% endblock %}
```
yanksyoon, line 6: Here as well :D

yanksyoon (approved):
LGTM!

gregory-schiano (approved):
LGTM

templates/haproxy_ingress.cfg.j2
Thanhphan1147: thanks, will add it in a follow-up PR.

templates/haproxy_legacy.cfg.j2
Thanhphan1147: thanks, will add it in a follow-up PR.

PR #29, opened 2024-10-23
update test and add logic to open ports

Thanhphan1147:
We need to manage the ports to be opened / closed based on the requested haproxy configuration. The opening/closing of ports will be done before applying the new haproxy config.

- in ingress mode, port 80 and 443 will be opened and all other ports closed
- in legacy mode, after generating the service definition, each port specified will be opened, all other ports are closed.
- in default mode, only port 80 will be opened. 

Exceptions related to opening/closing port will flow to the charm level and put the charm in an error state, therefore no validation is done (mainly regarding the legacy ports which can request any port value)
### Checklist


Thanhphan1147:
Labeling this to `complex` as we need to discuss whether to do validation for legacy ports, or let the exception flows.

Hi @gregory-schiano, can you take a look at this PR and review it when you have the time? Thanks!

src/charm.py
```diff
@@ -232,14 +233,21 @@ def _reconcile(self) -> None:
...
                 )
                 tls_information = TLSInformation.from_charm(self, self.certificates)
+                self.unit.set_ports(80, 443)
                 self.haproxy_service.reconcile_ingress(
                     config, ingress_requirers_information, tls_information.external_hostname
                 )
             case ProxyMode.LEGACY:
+                required_ports = set(
+                    Port(protocol="tcp", port=service["service_port"])
+                    for service in self.reverseproxy_requirer.get_services_definition().values()
```
nrobinaubertin: It would be comforting to add a type for the set (`set[int]`) and/or for the return value of `self.reverseproxy_requirer.get_services_definition()` to make sure that we're handling the right data. Is that possible ?

Thanhphan1147: I added `set[Port]` as the type for the `required_ports` variable so that it's clear what should be in it, is it ok?

nrobinaubertin: hm, sorry, I can't see it. Where did you add it ?

weiiwang01 (dismissed):

src/charm.py
Thanhphan1147: oops, pushed it to the wrong PR, sorry. I just correctly updated it

nrobinaubertin (dismissed):

weiiwang01 (dismissed):

src/charm.py
```diff
@@ -64,6 +65,18 @@ class ProxyMode(StrEnum):
     INVALID = "invalid"
 
 
+def validate_port(port: int) -> bool:
```
jdkandersson: Could this be a private function?

src/charm.py
```diff
@@ -232,14 +245,25 @@ def _reconcile(self) -> None:
...
+                self.unit.set_ports(80, 443)
                 self.haproxy_service.reconcile_ingress(
                     config, ingress_requirers_information, tls_information.external_hostname
                 )
             case ProxyMode.LEGACY:
+                required_ports: set[Port] = set()
+                for service in self.reverseproxy_requirer.get_services_definition().values():
+                    port = service["service_port"]
+                    if not validate_port(port):
+                        logger.error("Requested port: %s is not a valid tcp port. Skipping", port)
```
jdkandersson, line 202: If it is logged as an error, shouldn't the check go into blocked state?

jdkandersson: In terms of the logic, we are continuing execution which indicates that it is not an error for a port to be invalid. It would be worthwhile to think about whether this is an error scenario, what the user would be expecting and so on

Thanhphan1147: I implemented it like this because I was thinking about cases where haproxy can be related to multiple backend applications, and I don't think we should go into error state because one of the requirers has an invalid configuration, rather we can inform the operator via log messages.

Also it's not really an error if we're not opening the ports, as it doesn't affect the charm and I think a backend requesting an invalid port will not work regardless of the ingress implementation. I used `logger.error` here just to make the message visible when the operator looks at the logs.

src/charm.py
Thanhphan1147: makes sense, I updated it. Thanks!

src/charm.py
jdkandersson: Hmm, I think the charm should still go into blocked state. The workload should still serve for all the backends it can, it should be visible to the operator that there is something wrong. The operator is unlikely to go into the logs unless they know to take a look. What do you think?

weiiwang01 (dismissed):

src/charm.py
Thanhphan1147: Gotcha, I added logic to make the charm go into blocked state with a message if there's any service that requests an invalid port.

src/charm.py
```diff
@@ -176,14 +189,33 @@ def _reconcile(self) -> None:
...
+                required_ports: set[Port] = set()
+                for service in self.reverseproxy_requirer.get_services_definition().values():
+                    port = service["service_port"]
+                    if not _validate_port(port):
+                        logger.error("Requested port: %s is not a valid tcp port. Skipping", port)
+                        legacy_invalid_requested_port.append(f"{service['service_name']:{port}}")
+                        continue
+                    required_ports.add(Port(protocol="tcp", port=port))
+
+                if legacy_invalid_requested_port:
```
amandahla, line 207: I saw the discussion in the previous comment but I wonder if this port was passed via relation, shouldn't it emit some kind of event to the require warning about it instead of blocking the charm?
Example:
https://github.com/canonical/prometheus-k8s-operator/blob/62cd2d9b8c3a528910ce12c553209817e1b7a889/lib/charms/prometheus_k8s/v1/prometheus_remote_write.py#L114

src/charm.py
```diff
@@ -63,6 +64,18 @@ class ProxyMode(StrEnum):
     INVALID = "invalid"
 
 
+def _validate_port(port: int) -> bool:
```
amandahla, line 67: Out of scope but it would be nice if libraries validate data before adding to the relation so provider/requirer don't need to do this.

weiiwang01 (approved):

src/charm.py
Thanhphan1147: The legacy relation doesn't  have any data validation and we're not trying to bring it up to date, it's just here to support drop-in replacement. So we decided to try and at least do some data validation for the legacy interface but anyone using this charm should use either the `ingress` or `haproxy-route` interface rather than this legacy one.

src/charm.py
Thanhphan1147: There's currently no standardised interface for the `reverseproxy` relation and afaik the charms using this legacy interface is quite old and probably won't be updated in the future. So I think it's better to block in on our side since even if we pass the information over the relation data there's no guarantee that the other charm can handle it

amandahla (approved):

PR #50, opened 2025-01-10
ISD-2586 Support for ha in active/passive mode

Thanhphan1147:
Spec: ISD-196 (internal)

This PR adds support for the `ha` integration to be used with the `hacluster` charm to provide high-availability support through pacemaker/corosync, along with unit and integration tests.

Supported workflow:
```
juju deploy haproxy -n 3
juju deploy hacluster
juju config haproxy vip=<vip>
juju relate haproxy hacluster
```

### Checklist


requirements.txt
```diff
@@ -4,3 +4,4 @@ jsonschema==4.23.0
 ops==2.17.1
 pydantic==2.10.3
 cosl==0.0.47
+interface_hacluster @ git+https://github.com/charmed-kubernetes/charm-interface-hacluster@main
```
amandahla: Will Renovate work with this kind of requirement?

src/charm.py
```diff
@@ -253,6 +293,24 @@ def _validate_state(self) -> ProxyMode:
...
+        """Update ha configuration.
+
+        Args:
+            ha_information: HAInformation charm state component.
+        """
+        if not ha_information.integration_ready:
+            logger.info("ha integration is not ready, skipping.")
+            return
+        self.hacluster.add_vip(self.app.name, str(ha_information.vip))
+        self.hacluster.add_systemd_service(f"{self.app.name}-{HAPROXY_SERVICE}", HAPROXY_SERVICE)
```
amandahla, line 323: Is there a scenario where this service should be disabled?

tests/integration/test_ha.py
```diff
@@ -0,0 +1,35 @@
...
+    vip = await get_unit_ip_address(application)
+
+    await application.set_config({"vip": str(vip)})
+    await application.model.add_relation(f"{application.name}:ha", f"{hacluster.name}:ha")
+    await application.model.wait_for_idle(
+        apps=[application.name, hacluster.name],
+        idle_period=30,
+        status="active",
+    )
+    for unit in application.units:
```
amandahla: Would make sense adding a request too?

requirements.txt
gregory-schiano: At least using a tag to avoid using main branch

tests/integration/test_ha.py
gregory-schiano, line 1: We need a non happy-case scenario by killing the unit that has the VIP, and ensure a request still works.
I would also add the blocked state case where the integration is set but VIP is not defined

tests/integration/test_ha.py
Thanhphan1147: Greg made the same comments, I updated this test to do a GET request and also do another GET request after removing a machine

src/charm.py
Thanhphan1147: Since haproxy is stateless I decided to keep the service running even on passive units after discussing it with Greg

tests/integration/test_ha.py
Thanhphan1147: Added the logic of removing the machine with the vip to the current test

requirements.txt
Thanhphan1147: updated this to use a tag

amandahla (dismissed):

tests/integration/test_ha.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+async def test_ha(application: Application, hacluster: Application):
+    """
+    arrange: deploy the chrony charm.
+    act: request chrony_exporter metrics endpoint.
+    assert: confirm that metrics are scraped.
+    """
+    await application.add_unit(count=3)
+    await application.model.wait_for_idle(
```
arturo-seijas, line 18: I'd remove this since it might point us to race conditions when the async operations do not finish in the same order

Thanhphan1147: Even with `await`?

tox.ini
```diff
@@ -103,7 +103,7 @@ deps =
     protobuf==3.20.3
     pydantic
     pytest
-    juju==3.3.1.1
+    juju==3.6.1.0
     pytest-operator
     pytest-asyncio
     websockets<14.0 # https://github.com/juju/python-libjuju/issues/1184
```
arturo-seijas, line 109: Off-topic: Hasn't this been fixed already?

arturo-seijas (approved):

amandahla (approved):

tox.ini
Thanhphan1147: I'm not sure about that, but I also don't mind pinning pylibjuju to a more recent version.

tests/integration/test_ha.py
Thanhphan1147: I'm marking this as resolved for now, if you think we need to discuss about it more I'll move it to an internal thread!

PR #67, opened 2025-02-11
Improvements to handling certs, update template to handle ipv6

Thanhphan1147:
Update the exception handling logic and removed the unnecessary `block_on_tls_not_ready` parameter as we always want to block on tls-related events if tls is not ready to be handled.

### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -38,7 +38,7 @@ from http_interface import (
 from state.config import CharmConfig
 from state.ha import HACLUSTER_INTEGRATION, HAPROXY_PEER_INTEGRATION, HAInformation
 from state.ingress import IngressRequirersInformation
-from state.tls import TLSInformation, TLSNotReadyError
+from state.tls import TLSInformation
 from state.validation import validate_config_and_tls
 from tls_relation import TLSRelationService
 
@@ -90,6 +90,9 @@ class HAProxyCharm(ops.CharmBase):
         """
         super().__init__(*args)
         self.haproxy_service = HAProxyService()
+        # Order is important here as we want _ensure_tls to check if the hostname is configured
+        self.framework.observe(self.on[TLS_CERT_RELATION].relation_created, self._ensure_tls)
+        self.framework.observe(self.on[TLS_CERT_RELATION].relation_changed, self._ensure_tls)
         self.certificates = TLSCertificatesRequiresV4(
             charm=self,
             relationship_name=TLS_CERT_RELATION,
@@ -97,6 +100,7 @@ class HAProxyCharm(ops.CharmBase):
             refresh_events=[self.on.config_changed],
             mode=Mode.UNIT,
         )
+
         self._tls = TLSRelationService(self.model, self.certificates)
         self._ingress_provider = IngressPerAppProvider(charm=self, relation_name=INGRESS_RELATION)
         self.reverseproxy_requirer = HTTPRequirer(self, REVERSE_PROXY_RELATION)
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -136,16 +140,14 @@ def _on_install(self, _: typing.Any) -> None:
         self.haproxy_service.install()
         self.unit.status = ops.MaintenanceStatus("Waiting for haproxy to be configured.")
 
-    @validate_config_and_tls(defer=False, block_on_tls_not_ready=False)
```
arturo-seijas, line 139: At this point, shouldn't we merge `_on_config_changed`, `_on_certificate_available`, `_on_http_backend_available`, `_on_http_backend_removed` and all other `reconcile` wrappers for readability?

src/charm.py
```diff
@@ -169,34 +171,38 @@ def _on_get_certificate_action(self, event: ActionEvent) -> None:
...
-
         proxy_mode = self._validate_state()
         if proxy_mode == ProxyMode.INVALID:
             # We don't raise any exception/set status here as it should already be handled
             # by the _validate_state method
             return
 
+        if self.certificates._tls_relation_created():  # pylint: disable=protected-access
+            # Reconcile certificates in case the certificates relation is present
+            tls_information = TLSInformation.from_charm(self, self.certificates)
```
arturo-seijas, line 194: This is already available in a charm attr. See https://github.com/canonical/haproxy-operator/blob/8e83e8c6cdf4f38cd9e50b800d8621365007ad4d/src/charm.py#L194

src/charm.py
```diff
@@ -242,23 +248,23 @@ def _get_certificate_requests(self) -> typing.List[CertificateRequestAttributes]
...
         self._reconcile()
-        tls_information = TLSInformation.from_charm(self, self.certificates)
-        integration_data = self._ingress_provider.get_data(event.relation)
-        path_prefix = f"{integration_data.app.model}-{integration_data.app.name}"
-
-        self._ingress_provider.publish_url(
-            event.relation, f"https://{tls_information.external_hostname}/{path_prefix}/"
-        )
+        if self.unit.is_leader():
+            tls_information = TLSInformation.from_charm(self, self.certificates)
```
arturo-seijas, line 260: Isn't this already available in a charm attr? See https://github.com/canonical/haproxy-operator/blob/8e83e8c6cdf4f38cd9e50b800d8621365007ad4d/src/charm.py#L194

Thanhphan1147: I don't quite understand, can you elaborate on your comment?

The `tls_information` is there so that I can get the configured `external-hostname` to include in the ingress URL sent to the requirer application, I can use `charm.config.get` too but I think I prefer using a charmstate for it.

src/charm.py
Thanhphan1147: Makes sense, I'll do it in a follow-up PR.

src/charm.py
arturo-seijas: Nvm, this is cool :)

arturo-seijas (approved):

src/charm.py
```diff
@@ -90,13 +90,17 @@ def __init__(self, *args: typing.Any):
         """
         super().__init__(*args)
         self.haproxy_service = HAProxyService()
+        # Order is important here as we want _ensure_tls to check if the hostname is configured
+        self.framework.observe(self.on[TLS_CERT_RELATION].relation_created, self._ensure_tls)
+        self.framework.observe(self.on[TLS_CERT_RELATION].relation_changed, self._ensure_tls)
```
amandahla, line 95: Is it ok if is called twice? After relation created, a relation changed is emitted too.

src/charm.py
```diff
@@ -90,13 +90,17 @@ def __init__(self, *args: typing.Any):
         """
         super().__init__(*args)
         self.haproxy_service = HAProxyService()
+        # Order is important here as we want _ensure_tls to check if the hostname is configured
+        self.framework.observe(self.on[TLS_CERT_RELATION].relation_created, self._ensure_tls)
```
amandahla, line 94: btw would make sense to add a handler to the relation broken?

src/charm.py
Thanhphan1147: I believe that's fine since the handler is only checking if the charm is ready to handle TLS ( and defer it if that's not the case)

src/charm.py
Thanhphan1147: I don't think we need to do verification for `relation-broken` events ( the tls v4 library doesn't hook this event either ) Is there a specific scenario that you have in mind?

amandahla: I was thinking about wanting only http but now I remember that is required. So never mind ;)

amandahla (approved):

PR #82, opened 2025-03-04
ISD-3043 Render haproxy config for haproxy route

Thanhphan1147:
This PR adds the logic to render the haproxy config from the haproxy-route relation data. The structured data obtained from the relation is processed into a haproxy-specific structure and rendered with a custom j2 template.

Also changed some `logger.exception` lines to `logger.error` to reduce noise.

Added integration test for haproxy -route 

### Checklist


src/charm.py
```diff
@@ -278,10 +296,24 @@ def _validate_state(self) -> ProxyMode:
...
+                    related
+                    for related in [
+                        is_ingress_related,
+                        is_legacy_related,
+                        is_haproxy_route_related,
+                    ]
+                    if related
+                ]
+            )
+            > 1
```
gregory-schiano: Sum the bool since bool is a subclass of int

src/charm.py
```diff
@@ -278,10 +296,24 @@ def _validate_state(self) -> ProxyMode:
...
+        ):
+            logger.error("Only one relation type is supported at a time.")
+            self.unit.status = ops.BlockedStatus("Only one relation type is supported at a time.")
```
gregory-schiano: Be more specific, probably specify the relations which are set

src/charm.py
```diff
@@ -324,6 +360,13 @@ def _reconcile_ha(self, ha_information: HAInformation) -> None:
...
 
+    @validate_config_and_tls(defer=False)
+    def _configure_haproxy_route(self, _: HAServiceReadyEvent) -> None:
+        """Handle the ha-ready event."""
+        data = self.haproxy_route_provider.get_data(self.haproxy_route_provider.relations)
+        # This is temporary as the logic to generate the haproxy config will be added later.
+        logger.debug("Aggregated requirer data: %s", data)
+
     @validate_config_and_tls(defer=True)
     def _ensure_tls(self, _: ops.EventBase) -> None:
```
gregory-schiano, line 397: Unused method

src/charm.py
```diff
@@ -228,6 +240,12 @@ def _reconcile(self) -> None:
                 self.haproxy_service.reconcile_legacy(
                     config, self.reverseproxy_requirer.get_services()
                 )
+            case ProxyMode.HAPROXY_ROUTE:
```
gregory-schiano, line 250: Suggested refactoring:
- Do no validate TLS in _validate_state and rename the method to "get_mode", "define_mode"
- Either execute the TLS required code in each case statement (adding the if _tls_relation_created for legacy)
- Or Change the previous if statement to be more explicit (if INGRESS or HAPROXY_ROUTE or (LEGACY & _tls_relation_created)
Either way, the validation is done when calling TLSInformation.from_charm (we'll rely on that)

src/state/haproxy_route.py
```diff
@@ -0,0 +1,304 @@
...
+        """
+        if self.application_data.load_balancing.algorithm == LoadBalancingAlgorithm.COOKIE:
+            # The library needs to ensure that if algorithm == cookie
+            # then the cookie attribute must be not none
+            return f"hash req.cookie({cast(str, self.application_data.load_balancing.cookie)})"
+        return str(self.application_data.load_balancing.algorithm)
+
+    @property
+    def rewrite_configurations(self) -> list[str]:
+        """Build the rewrite configurations.
```
gregory-schiano, line 149: Give an example of the string rendered for each property

src/state/haproxy_route.py
```diff
@@ -0,0 +1,304 @@
...
+                )
+                backends.append(backend)
+
+            # Fetch address of all units for `peer:` section
+            peers: list[str] = []
+            unit_address = get_unit_address(charm)
+            if not unit_address:
+                raise HaproxyRouteIntegrationDataValidationError("Couldn't get the unit address.")
+            peers.append(unit_address)
+            peers.extend(get_peer_units_address(charm))
```
gregory-schiano: Since the charm is only used to get the unit and peer IP addresses, pass these info from the charm, this avoids depending on the whole charm

src/state/haproxy_route.py
```diff
@@ -0,0 +1,304 @@
...
+
+    Attrs:
+        backends: The list of backends each corresponds to a requirer application.
+        stick_table_entries: List of stick table entries in the haproxy "peer" section.
+        peers: List of IP address of haproxy peer units.
+    """
+
+    backends: list[HAProxyRouteBackend]
+    stick_table_entries: list[str]
+    peers: list[AnyHttpUrl]
```
gregory-schiano: Should probably be IPvAnyAddress type

src/state/haproxy_route.py
```diff
@@ -0,0 +1,263 @@
...
+                    )
+                backend_names.add(requirer.application_data.service)
+                stick_table_entries.append(f"{requirer.application_data.service}_rate_limit")
+
+                backend = HAProxyRouteBackend(
+                    application_data=requirer.application_data,
+                    servers=get_servers_definition_from_requirer_data(requirer),
+                    external_hostname=tls_information.external_hostname,
+                )
+                backends.append(backend)
```
gregory-schiano, line 223: We probably need to implement the backend ordering logic here, based on path max depth

src/state/haproxy_route.py
```diff
@@ -0,0 +1,263 @@
...
+    Returns:
+        list[HAProxyRouteServer]: List of server definitions.
+    """
+    servers: list[HAProxyRouteServer] = []
+    for i, unit_data in enumerate(requirer.units_data):
+        for port in requirer.application_data.ports:
+            servers.append(
+                HAProxyRouteServer(
+                    server_name=f"{requirer.application_data.service}_{port}_{i}",
+                    host=unit_data.host,
```
gregory-schiano: Due to haproxy behavior in resolving, we should probably use IP addresses here as using hostname has side effects (cache, timeout of DNS resolution)

But as a consequence, the healthcheck won't used any "Host" header, which might fail in some  circumstances.

templates/haproxy_route.cfg.j2
```diff
@@ -0,0 +1,48 @@
...
+
+{% for backend in backends %}
+backend {{ backend.backend_name }}
+    balance {{ backend.load_balancing_configuration }}
+    timeout server {{ backend.application_data.timeout.server }}
+    timeout connect {{ backend.application_data.timeout.connect }}
+    timeout queue {{ backend.application_data.timeout.queue }}
+{% if backend.application_data.check.path %}
+    option httpchk GET {{ backend.application_data.check.path }}
+{% endif %}
```
gregory-schiano: Duplicated

templates/haproxy_route.cfg.j2
```diff
@@ -0,0 +1,48 @@
...
+
+{% if backend.application_data.check.path %}
+    option httpchk GET {{ backend.application_data.check.path }}
```
gregory-schiano: Add the host header to the check query (pick the first in hostname_acls)

Thanhphan1147: Added, thanks!

templates/haproxy_route.cfg.j2
Thanhphan1147: removed!

src/state/haproxy_route.py
Thanhphan1147: After discussing this internally, I'll remove the possibility of passing a hostname for now. As a result the requirer can only send IP addresses for each unit, this will be validated with pydantic's `IPvAnyAddress`

src/state/haproxy_route.py
Thanhphan1147: Updated!

src/state/haproxy_route.py
Thanhphan1147: I moved this outside the module and into `charm.py`

src/state/haproxy_route.py
Thanhphan1147: Updated the docstring with examples.

src/charm.py
Thanhphan1147: Updated, I also added cert rendering for default mode.

src/charm.py
Thanhphan1147: I checked and this is actually used due to how the v4 certificates library is implemented. We need to keep this here for now.

src/charm.py
Thanhphan1147: Updated this to give a more detailed message

src/charm.py
Thanhphan1147: Thanks, updated!

.github/workflows/integration_test.yaml
```diff
@@ -12,4 +12,8 @@ jobs:
       juju-channel: 3/stable
       self-hosted-runner: true
       charmcraft-channel: latest/edge
-      modules: '["test_action.py", "test_config.py", "test_charm.py", "test_http_interface.py", "test_cos.py", "test_ingress.py", "test_ha.py"]'
+      modules: '["test_action.py", "test_config.py", "test_charm.py", "test_http_interface.py", "test_cos.py", "test_ingress.py", "test_ha.py", "test_haproxy_route.py"]'
+      pre-run-script: |
+        -c "curl -L https://github.com/canonical/haproxy-operator/releases/download/rev136/any-charm_ubuntu-22.04-amd64.charm -o ${GITHUB_WORKSPACE}/any-charm_ubuntu-22.04-amd64.charm"
+      extra-arguments: |
+        --any-charm-file=${GITHUB_WORKSPACE}/any-charm_ubuntu-22.04-amd64.charm
```
arturo-seijas: missing blank line at EOF

.github/workflows/integration_test.yaml
```diff
@@ -12,4 +12,8 @@ jobs:
       juju-channel: 3/stable
       self-hosted-runner: true
       charmcraft-channel: latest/edge
-      modules: '["test_action.py", "test_config.py", "test_charm.py", "test_http_interface.py", "test_cos.py", "test_ingress.py", "test_ha.py"]'
+      modules: '["test_action.py", "test_config.py", "test_charm.py", "test_http_interface.py", "test_cos.py", "test_ingress.py", "test_ha.py", "test_haproxy_route.py"]'
```
arturo-seijas: Can you order this alphabetically? IT'll be easy to see if any test is missing or fine one if we want to remove it, etc

.github/workflows/integration_test.yaml
```diff
@@ -12,4 +12,8 @@ jobs:
...
+      pre-run-script: |
+        -c "curl -L https://github.com/canonical/haproxy-operator/releases/download/rev136/any-charm_ubuntu-22.04-amd64.charm -o ${GITHUB_WORKSPACE}/any-charm_ubuntu-22.04-amd64.charm"
```
arturo-seijas: Can't we fetch any-charm from charmhub?

lib/charms/haproxy/v0/haproxy_route.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, *args):
     # There are 2 ways you can use the requirer implementation:
     # 1. To initialize the requirer with parameters:
     self.haproxy_route_requirer = HaproxyRouteRequirer(self,
-        host=<required>,
+        address=<required>,
```
arturo-seijas, line 34: You have to increase the lib version

src/charm.py
```diff
@@ -197,17 +200,15 @@ def _reconcile(self) -> None:
...
-            tls_information = TLSInformation.from_charm(self, self.certificates)
-            self._tls.certificate_available(tls_information)
-
         ha_information = HAInformation.from_charm(self)
         self._reconcile_ha(ha_information)
 
         config = CharmConfig.from_charm(self)
         match proxy_mode:
             case ProxyMode.INGRESS:
+                tls_information = TLSInformation.from_charm(self, self.certificates)
```
arturo-seijas, line 214: As far as I can see you are executing these two lines on every case
```
 tls_information = TLSInformation.from_charm(self, self.certificates)
self._tls.certificate_available(tls_information)
```
Can you move them out of the match-case?

src/charm.py
```diff
@@ -338,12 +390,46 @@ def _configure_haproxy_route(self, _: HAServiceReadyEvent) -> None:
...
+    def _get_peer_units_address(self) -> list[str]:
+        """Get address of peer units.
+
+        Returns:
+            liststr]: The list of peer units address.
+        """
+        peer_units_address: list[str] = []
+        if haproxy_peer_integration := self.model.get_relation(HAPROXY_PEER_INTEGRATION):
+            for unit in haproxy_peer_integration.units:
+                if unit != self.unit:
```
arturo-seijas, line 415: I would remove this condition and return the list including all units, which is what you end up using in line 258. Also, consider if you still need `self.unit`

src/charm.py
```diff
@@ -338,12 +390,46 @@ def _configure_haproxy_route(self, _: HAServiceReadyEvent) -> None:
...
+                    if peer_unit_address := haproxy_peer_integration.data[unit].get(
+                        "private-address"
+                    ):
+                        peer_units_address.append(peer_unit_address)
+                    else:
+                        logger.error("Cannot get address for peer unit: %s. Skipping", unit)
```
arturo-seijas: Shouldn't this be a debug level log entry?

src/haproxy.py
```diff
@@ -157,7 +180,7 @@ def _reload_haproxy_service(self) -> None:
         try:
             systemd.service_reload(HAPROXY_SERVICE)
         except systemd.SystemdError as exc:
-            logger.exception("Failed reloading the haproxy service.")
+            logger.error("Failed reloading the haproxy service.")
```
arturo-seijas: This log is probably redundant

src/state/haproxy_route.py
```diff
@@ -0,0 +1,294 @@
...
+        server_name: The name of the unit with invalid characters replaced.
+        address: The IP address of the requirer unit.
+        port: The port that the requirer application wishes to be exposed.
+        check: Health check configuration.
+        maxconn: Maximum allowed connections before requests are queued.
+    """
+
+    server_name: str
+    address: IPvAnyAddress
+    port: int
```
arturo-seijas, line 50: Can you validate the range here?

src/state/haproxy_route.py
```diff
@@ -0,0 +1,294 @@
...
+    check: ServerHealthCheck
+    maxconn: Optional[int]
```
arturo-seijas, line 52: Same here. Positive int? > -1?

src/state/haproxy_route.py
```diff
@@ -0,0 +1,294 @@
...
+                backends.append(backend)
+
+            return HaproxyRouteRequirersInformation(
+                backends=backends,
+                stick_table_entries=stick_table_entries,
+                peers=list(map(lambda x: cast(IPvAnyAddress, x), peers)),
+            )
+        except DataValidationError as exc:
+            # This exception is only raised if the provider has "raise_on_validation_error" set
+            logger.error("Error validating requirer data: %s", str(exc))
```
arturo-seijas: This log is probably redundant

src/state/haproxy_route.py
```diff
@@ -0,0 +1,294 @@
...
+        """
+        requirers_paths: list[str] = []
+        requirers_hostnames: list[str] = []
+
+        for backend in self.backends:
+            requirers_paths.extend(backend.application_data.paths)
+            requirers_hostnames.extend(backend.hostname_acls)
+
+        if len(requirers_paths) != len(set(requirers_paths)):
+            logger.error(
```
arturo-seijas: If you don't plan to block under this scenario, this should probably be a warning

src/state/haproxy_route.py
```diff
@@ -0,0 +1,294 @@
...
+                (
+                    "Requirers defined path(s) that map to multiple backends."
+                    "This can cause unintended behaviours."
+                )
+            )
+
+        if len(requirers_hostnames) != len(set(requirers_hostnames)):
+            logger.error(
```
arturo-seijas: Same here

templates/haproxy_route.cfg.j2
```diff
@@ -0,0 +1,67 @@
...
+
+{% for rewrite_configuration in backend.rewrite_configurations %}
+    http-request {{ rewrite_configuration }}
+{% endfor %}
+{% for server in backend.servers %}
+    server {{ server.server_name }} {{ server.address }}:{{ server.port }} check inter {{ server.check.interval }}s rise {{ server.check.rise }} fall {{ server.check.fall }}
+{% endfor %}
+
+{% endfor %}
+{% endblock %}
```
arturo-seijas: Missing blank line at EOF

src/state/haproxy_route.py
Thanhphan1147: Thanks, I was thinking to use `error` here to highlight that this is quite important and the routing will probably get messed up. changed it to warning.

templates/haproxy_route.cfg.j2
Thanhphan1147: thanks, added!

src/charm.py
Thanhphan1147: I did it like this because in cases where you only have 1 unit the `haproxy-peers` relation will not exists, but we need the main unit IP address regardless to build the `peers` section of the config and raise an error if that's not available.

src/charm.py
Thanhphan1147: I had a discussion with greg about this and we agreed that writing the certifcates to fs explicitly in each case makes more sense mainly because we need to initialize `tls_information` locally in each match case anyways ( since  `tls_information` is required for `INGRESS` and `HAPROXY_ROUTE` but optional for `LEGACY` and `DEFAULT`.

If we moved this outside then we'd need to add more logic to manage the optional cases or re-initialzes `tls_information` for `INGRESS` and `HAPROXY_ROUTE`

lib/charms/haproxy/v0/haproxy_route.py
Thanhphan1147: Thanks, updated!

.github/workflows/integration_test.yaml
Thanhphan1147: thanks, updated to make it more readable and ordered them alphabetically

src/charm.py
arturo-seijas: I think it will be clearer if that logic was within a single function returning all the unit IPs regardless of whether the peer relation exists or not

src/state/haproxy_route.py
Thanhphan1147: This should already been validated by the `get_data` method of the Provider class in the library (I think the actual validation is also not there so I'll add it in :smile: ), do you think we should also do validation here? ( that'd mean do validation twice )

I implemented the state component with the main purpose to structure already validated data into haproxy-specific language to help with rendering

src/haproxy.py
Thanhphan1147: I put it here because I was trying to follow the pattern of logging every time I catch an exception, do you think that I should remove this line?

src/charm.py
Thanhphan1147: updated this to warning!

src/haproxy.py
arturo-seijas: I suspect you are already logging that exception when you catch it in the caller. That would probably lead to less code too

Thanhphan1147: makes sense, I'll remove it. Thanks!

src/charm.py
Thanhphan1147: got it, I moved the 2 logics into the same method.

src/state/haproxy_route.py
Thanhphan1147: removed!

.github/workflows/integration_test.yaml
Thanhphan1147: Yes, this was here while waiting for the latest revision to be published, I removed it.

.github/workflows/integration_test.yaml
Thanhphan1147: added a newline !

src/state/haproxy_route.py
arturo-seijas: Sounds good to me. Thanks for clarifying :)

src/state/haproxy_route.py
Thanhphan1147: same comments as below. we won't do validation here as the state component is mainly responsible for organizing data into haproxy language, while validation is done in the  `get_data` method of the library

arturo-seijas (approved):

arturo-seijas (approved):

.github/workflows/integration_test.yaml
```diff
@@ -12,4 +12,4 @@ jobs:
       juju-channel: 3/stable
       self-hosted-runner: true
       charmcraft-channel: latest/edge
-      modules: '["test_action.py", "test_config.py", "test_charm.py", "test_http_interface.py", "test_cos.py", "test_ingress.py", "test_ha.py"]'
+      modules: '["test_action.py","test_config.py","test_cos.py","test_charm.py","test_ha.py","test_haproxy_route.py","test_http_interface.py","test_ingress.py"]'
```
seb4stien: As the goal seems to be to sort them: test_charm should go before test_config ;)

src/charm.py
```diff
@@ -236,7 +247,37 @@ def _reconcile(self) -> None:
...
+                self.unit.set_ports(80, 443)
+                if self.unit.is_leader():
+                    for backend in haproxy_route_requirers_information.backends:
+                        relation = self.model.get_relation(
+                            HAPROXY_ROUTE_RELATION, backend.relation_id
+                        )
+                        if not relation:
+                            logger.error("Relation does not exist, skipping.")
+                        self.haproxy_route_provider.publish_proxied_endpoints(
+                            list(map(lambda x: f"https://{x}", backend.hostname_acls)), relation
```
seb4stien: [f"https://{x}" for x in backend.hostname_acls] may be a bit simpler.

Thanhphan1147: Ah indeed, I updated the code here, thanks.

tests/integration/haproxy_route_requirer_src.json
```diff
@@ -0,0 +1 @@
...s = str(bind_address)\n            else:\n                logger.error(\"No unit IP available.\")\n                raise DataValidationError(\"No unit IP available.\")\n        return RequirerUnitData(address=cast(IPvAnyAddress, address))\n\n    def get_proxied_endpoints(self) -> list[AnyHttpUrl]:\n        \"\"\"The full ingress URL to reach the current unit.\n\n        Returns:\n            The provider URL or None if the URL isn't available yet or is not valid.\n        \"\"\"\n        relation = self.relation\n        if not relation or not relation.app:\n            return []\n\n        # Fetch the provider's app databag\n        try:\n            databag = relation.data[relation.app]\n        except ModelError:\n            logger.exception(\"Error reading remote app data.\")\n            return []\n\n        if not databag:  # not ready yet\n            return []\n\n        try:\n            provider_data = cast(\n                HaproxyRouteProviderAppData, HaproxyRouteProviderAppData.load(databag)\n            )\n            return provider_data.endpoints\n        except DataValidationError:\n            logger.exception(\"Invalid provider url.\")\n            return []\n"}
```
seb4stien: Storing the json doesn't seem easy to maintain. Could you store the script in plain text and generate the config dynamically?
I think this is what is done in content-cache: https://github.com/canonical/content-cache-operator/blob/11ae0065f398d594917468fcec6b632069bdb05e/content-cache/tests/integration/helpers.py#L21

tests/integration/test_haproxy_route.py
```diff
@@ -0,0 +1,65 @@
...
+    configured_application_with_tls: Application,
+    haproxy_route_requirer: Application,
+):
+    """Deploy the charm with anycharm haproxy-route requirer that installs apache2.
+
+    Assert that the requirer endpoint is available.
+    """
+    application = configured_application_with_tls
+    action = await haproxy_route_requirer.units[0].run_action("rpc", method="start_server")
+    await action.wait()
```
seb4stien: I don't get the purpose of the run_action here. The fixture seems to return the app when it's "active". Why do you need to call "start_server"?
If something is required, shouldn't it be done in the fixture?

src/state/haproxy_route.py
```diff
@@ -0,0 +1,300 @@
...
+                backends.append(backend)
+
+            return HaproxyRouteRequirersInformation(
+                # Sort backend by the max depth of the required path.
+                # This is to ensure that backends with deeper path ACLs get routed first.
+                backends=sorted(
+                    backends,
+                    key=lambda backend: max(
+                        map(len, cast(HAProxyRouteBackend, backend).application_data.paths)
+                    ),
```
seb4stien: max(len(path) for path in cast(HAProxyRouteBackend, backend).application_data.paths) ?

And shouldn't the longest go first, so it should be "-max"?

src/state/haproxy_route.py
```diff
@@ -0,0 +1,300 @@
...
+                ),
+                stick_table_entries=stick_table_entries,
+                peers=list(map(lambda x: cast(IPvAnyAddress, x), peers)),
```
seb4stien: [cast(IPvAnyAddress, x) for x in peers] ;)

src/state/haproxy_route.py
```diff
@@ -0,0 +1,300 @@
...
+            Self: The validated model.
+        """
+        requirers_paths: list[str] = []
+        requirers_hostnames: list[str] = []
+
+        for backend in self.backends:
+            requirers_paths.extend(backend.application_data.paths)
+            requirers_hostnames.extend(backend.hostname_acls)
+
+        if len(requirers_paths) != len(set(requirers_paths)):
```
seb4stien, line 250: Here we catch exact matches. Don't we have a risk if we have overlaps betweens paths?
Something like:
- /patha
- /patha/andb

seb4stien (changes requested):
Left some comments.

templates/haproxy_route.cfg.j2
```diff
@@ -0,0 +1,67 @@
+{% extends 'haproxy.cfg.j2' %}
+{% block proxy_configuration %}
+frontend haproxy
+    mode http
+    bind [::]:80 v4v6
+    bind [::]:443 v4v6 ssl crt {{ haproxy_crt_dir }}
```
amandahla, line 6: Maybe this is out of context of this PR but just a heads-up: with IRC Bridge using ingress, I saw this bind 443 being enabled before having the certificates, resulting in an error.

src/state/haproxy_route.py
amandahla: sorted reverse=true?

amandahla: btw fancy ;-)
`key=lambda backend: max(
                        map(len, cast(HAProxyRouteBackend, backend).application_data.paths)
                    ),`

tests/integration/haproxy_route_requirer_src.json
Thanhphan1147: I stored is as a json because it'll also have a purpose in documentation to quickly deploy a requirer application since we don't really have machine requirer charms for ingress and haproxy-route yet. The idea is to write something like this in our docs:
```
juju deploy any-charm requirer --channel beta --config src-overwrite="$(curl https://raw.githubusercontent.com/canonical/haproxy-operator/refs/heads/main/tests/integration/haproxy_route_requirer.json)" --config python-packages="pydantic"
```

Thanhphan1147: Asking the user to generate the json with a script might be too complex, what do you think?

tests/integration/test_haproxy_route.py
Thanhphan1147: This action is used to install and start the apache2 server and setup the necessary paths. It's probably make more sense to move it to the fixture. I'll do that, thanks!

src/state/haproxy_route.py
Thanhphan1147: Removed the fancy and unnecssary one-liner and moved it to a separate method :)

src/state/haproxy_route.py
Thanhphan1147: update! thanks!

templates/haproxy_route.cfg.j2
Thanhphan1147: That's strange, haproxy should block until certificates are available. Let's move this discussion internally

tests/integration/haproxy_route_requirer_src.json
seb4stien: This is not what I meant ;)
Instead of having:
```
"src-overwrite": pathlib.Path(HAPROXY_ROUTE_REQUIRER_SRC).read_text(encoding="utf-8")
```
Having "any_charm.py" stored as a regular python file in the repo (in tests/integration/scripts) and use something like:
```
"src-overwrite": {
  "any-charm": pathlib.Path(path/to/your/python/scropt).read_text(encoding="utf-8")
 ...
}
```

Thanhphan1147: Ah i see, I think the problem with that is that any_charm.py depends on `any_charm_base` which doesn't exist in our charm repo ( it only exists in the any-charm unit once deployed )

```
from any_charm_base import AnyCharmBase
```
So if we put it in our code we'd need special rules to ignore the lint error for that file.

Also the json src-overwrite contains more than one file, aside from `any_charm.py` you also need the haproxy_route library and the `apt` library to install apache2, you'd need to get all 3 files together in a dict and **then** generate the json object:

```
print(
    json.dumps(
        {
            "any_charm.py": any_charm_py,
            "apt.py": pathlib.Path(
                "/home/ubuntu/Home/personal/haproxy_any_charm/ingress/apt.py"
            ).read_text(encoding="utf-8"),
            "haproxy_route.py": pathlib.Path(
                "/home/ubuntu/Home/Canonical/haproxy-operator/lib/charms/haproxy/v0/haproxy_route.py"
            ).read_text(encoding="utf-8"),
        }
    )
)
```

We can absolutey do this in the fixture but I was thinking that doing all of that would be too involved in a documentation context, so to serve both use-cases I decided to use the generated json.

tests/integration/test_haproxy_route.py
Thanhphan1147: moved the `start_server` to the fixture

src/state/haproxy_route.py
Thanhphan1147: Not really as we sort the backends based on path depth, so the backend that specify `/patha/andb` would get matched first and routed first, while other requests only matching /patha will be routed to the other backend.

.github/workflows/integration_test.yaml
Thanhphan1147: sorted, thanks!

src/charm.py
```diff
@@ -246,25 +251,33 @@ def _reconcile(self) -> None:
...
                 self.unit.set_ports(80, 443)
+                if self.unit.is_leader():
+                    for backend in haproxy_route_requirers_information.backends:
+                        relation = self.model.get_relation(
+                            HAPROXY_ROUTE_RELATION, backend.relation_id
+                        )
+                        if not relation:
+                            logger.error("Relation does not exist, skipping.")
+                        self.haproxy_route_provider.publish_proxied_endpoints(
+                            [f"https://{hostname}" for hostname in backend.hostname_acls], relation
```
gregory-schiano: We should provide the full urls including the path ACLs

src/state/haproxy_route.py
```diff
@@ -254,10 +286,27 @@ def get_servers_definition_from_requirer_data(
...
+    Args:
+        backend: haproxy-route backend.
+
+    Returns:
+        int: The max requested path depth
+    """
+    paths = backend.application_data.paths
+    if not paths:
+        return 1
+    return max(len(path) for path in paths)
```
gregory-schiano: You need to split on "/" to get the depth

tests/integration/haproxy_route_requirer_src.json
gregory-schiano: As discussed during our project meeting, we should try to put the files in a folder in this repo, and use the CI to generate the JSON file which will be put as an artefact which can be downloaded

templates/haproxy_route.cfg.j2
amandahla: I'll mark this as resolved until we find exactly how to reproduce.

src/state/haproxy_route.py
amandahla: oh I liked the fancy one but moving is ok too ;-)

src/state/haproxy_route.py
```diff
@@ -0,0 +1,312 @@
...
+        stick_table_entries: List of stick table entries in the haproxy "peer" section.
+        peers: List of IP address of haproxy peer units.
+    """
+
+    backends: list[HAProxyRouteBackend]
+    stick_table_entries: list[str]
+    peers: list[IPvAnyAddress]
+
+    @classmethod
+    def from_provider(
```
amandahla, line 185: Could this be validated via unit test?

src/state/haproxy_route.py
Thanhphan1147: added, thanks!

src/charm.py
Thanhphan1147: updated to send the full urls

src/state/haproxy_route.py
Thanhphan1147: yes, added unit tests, thanks!

tests/integration/haproxy_route_requirer_src.json
Thanhphan1147: added this logic and removed the json file

src/state/haproxy_route.py
Thanhphan1147: Added ordering logic!

.github/workflows/publish_charm.yaml
```diff
@@ -16,3 +16,31 @@ jobs:
     secrets: inherit
     with:
       channel: 2.8/edge
+
+  generate-anycharm-json-artifact:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: |
+          python3 - <<'____HERE' > /tmp/haproxy_route_requirer_src.json
```
seb4stien: "EOF" is more standard (instead of '____HERE'), you don't need the quotes.

Thanhphan1147: Updated it to 'EOF', I need the quotes here to avoid problems with the quotes in the python script itself I think, otherwise I'd need to escape them.

tox.ini
```diff
@@ -61,7 +61,8 @@ commands =
     # codespell {[vars]lib_path}
     codespell {toxinidir} --skip {toxinidir}/.git --skip {toxinidir}/.tox \
       --skip {toxinidir}/build --skip {toxinidir}/lib --skip {toxinidir}/venv \
-      --skip {toxinidir}/.mypy_cache --skip {toxinidir}/icon.svg --skip {toxinidir}/src/grafana_dashboards/haproxy.json 
+      --skip {toxinidir}/.mypy_cache --skip {toxinidir}/icon.svg --skip {toxinidir}/src/grafana_dashboards/haproxy.json \
+      --skip {toxinidir}/tests/integration/haproxy_route_requirer_src.json
```
seb4stien, line 65: This is a leftover right?  You can remove the reference to haproxy_route_requirer_src now?

.github/workflows/publish_charm.yaml
```diff
@@ -16,3 +16,31 @@ jobs:
     secrets: inherit
     with:
       channel: 2.8/edge
+
+  generate-anycharm-json-artifact:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: |
+          python3 - <<'EOF' > /tmp/haproxy_route_requirer_src.json
```
seb4stien, line 25:
```suggestion
          python3 - << EOF > /tmp/haproxy_route_requirer_src.json
```

seb4stien (approved):

tox.ini
Thanhphan1147: yep, will update it in a follow-up PR

PR #111, opened 2025-05-21
Temporary disable sendhdr for httpchk

Thanhphan1147:
Some application ( like pollen ) doesn't care about the generated ingress URL and thus does not expect healthcheck requests to come with a header. In those cases haproxy will refuse to proxy incoming requests and replied with 503.

Until we find a more suitable solution i suggest we disable sending host headers in httpchk.

This PR also update the libpatch version to properly align with charmhub

### Checklist

### DIFF
--- a/templates/haproxy_route.cfg.j2
+++ b/templates/haproxy_route.cfg.j2
@@ -38,7 +38,10 @@ backend {{ backend.backend_name }}
     timeout queue {{ backend.application_data.timeout.queue }}
 
     option httpchk {% if backend.application_data.check.path %}GET {{ backend.application_data.check.path }}{% endif %}
-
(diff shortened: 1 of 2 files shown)


templates/haproxy_route.cfg.j2
```diff
@@ -39,8 +39,6 @@ backend {{ backend.backend_name }}
 
     option httpchk {% if backend.application_data.check.path %}GET {{ backend.application_data.check.path }}{% endif %}
 
-    http-check send hdr Host {{ backend.hostname_acls[0] }}
```
seb4stien, line 42: Shouldn't we comment it and leave a comment on the "why" as we have forgotten in a few weeks/months?

Thanhphan1147: thanks, I updated that part

seb4stien (approved):

PR #118, opened 2025-05-23
Validate requirer relation data object to check for invalid characters.

Thanhphan1147:
Validation is done on both the requirer and provider side. On the requirer an exception will be raised while on the provider the related service will be ignored.

### Checklist


lib/charms/haproxy/v0/haproxy_route.py
```diff
@@ -526,6 +528,46 @@ def validate_rewrites(cls, rewrites: list[RewriteConfiguration]) -> list[Rewrite
...
+        data: The object to sanitize.
+
+    Returns:
+        The sanitized object.
+    """
+    if hasattr(data, "__iter__"):
+        return [sanitize_data(item) for item in data]
+
+    if isinstance(data, dict):
+        for attr, value in data.items():
```
seb4stien: Shouldn't you just `data[attr] = sanitize_data(value)`  and move the isinstance(value, str) test at the beginning of the function?
I feel we have the logic in two places with the current approach.

lib/charms/haproxy/v0/haproxy_route.py
```diff
@@ -526,6 +528,46 @@ def validate_rewrites(cls, rewrites: list[RewriteConfiguration]) -> list[Rewrite
...
+            if isinstance(value, str):
+                data[attr] = value.translate(TRANSLATION_TABLE)
```
seb4stien: I wonder if we should not log something if "value.translate != value"? It could help troubleshooting.

lib/charms/haproxy/v0/haproxy_route.py
Thanhphan1147: I discussed with @gregory-schiano and we agreed that simply validating the data and raising an error if we find any invalid characters should be enough so that we don't have write any complex logic for sanitization.

lib/charms/haproxy/v0/haproxy_route.py
Thanhphan1147: I changed the approach as mentioned above!

seb4stien:
Shouldn't we have at least one test with a bad character?

seb4stien (approved):

lib/charms/haproxy/v0/haproxy_route.py
```diff
@@ -142,10 +143,31 @@ def _on_haproxy_route_data_available(self, event: EventBase) -> None:
...
+    """Validate if value contains invalid haproxy config characters.
+
+    Args:
+        value: The value to validate.
+
+    Raises:
+        ValueError: When value contains invalid characters.
+
+    Returns:
+        str: The validated value.
```
arturo-seijas: s/ str: The validated value./  The validated value.

lib/charms/haproxy/v0/haproxy_route.py
```diff
@@ -142,10 +143,31 @@ def _on_haproxy_route_data_available(self, event: EventBase) -> None:
...
+    """
+    if value is None:
+        return value
+
+    if [c for c in value if c in HAPROXY_CONFIG_INVALID_CHARACTERS]:
```
arturo-seijas: I think "char" is more readable than "c"

lib/charms/haproxy/v0/haproxy_route.py
```diff
@@ -142,10 +143,31 @@ def _on_haproxy_route_data_available(self, event: EventBase) -> None:
...
+        raise ValueError("Relation data contains invalid character(s)")
```
arturo-seijas: Can you to the message here the value that triggered the exception?

Thanhphan1147: makes sense, updated!

lib/charms/haproxy/v0/haproxy_route.py
Thanhphan1147: makes sense, I updated it!

lib/charms/haproxy/v0/haproxy_route.py
Thanhphan1147: Thanks, updated!

arturo-seijas (approved):

PR #136, opened 2025-06-20
pin haproxy version to 2.8.5-1ubuntu3.3

Thanhphan1147:
With `2.8.5-1ubuntu3.3` available, running `apt install -y haproxy=2.8.5-1ubuntu3` will result in the following error:

```
charms.operator_libs_linux.v0.apt.PackageError: Could not install package(s) ['haproxy=2.8.5-1ubuntu3']: E: Packages were downgraded and -y was used without --allow-downgrades.
```

Also added a call to `apt-mark hold haproxy` to pin the haproxy version and prevent automatic updates. 

### Checklist

### DIFF
--- a/src/haproxy.py
+++ b/src/haproxy.py
@@ -16,7 +16,7 @@ from state.config import CharmConfig
 from state.haproxy_route import HaproxyRouteRequirersInformation
 from state.ingress import IngressRequirersInformation
 
-APT_PACKAGE_VERSION = "2.8.5-1ubuntu3"
+APT_PACKAGE_VERSION = "2.8.5-1ubuntu3.3"
 APT_PACKAGE_NAME = "haproxy"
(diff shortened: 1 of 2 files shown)


src/haproxy.py
```diff
@@ -65,8 +65,9 @@ def install(self) -> None:
         Raises:
             RuntimeError: If the service is not running after installation.
         """
-        apt.update()
```
swetha1654, line 68: why are we removing apt.update? isn't it required for the APT_PACKAGE to be available?

swetha1654 (approved):

src/haproxy.py
Thanhphan1147: yep, I added the update in the `add_package` call itself just below with `update_cache=True`

seb4stien (approved):

PR #144, opened 2025-07-01
ISD-3795 update certificate handling, update tests, request cert for haproxy-route backends using subdomains

Thanhphan1147:
Currently haproxy only request cert for the external-hostname config, if a haproxy-route requirer charm requested a subdomain the resulting fqdn should also have a certificate.

In this PR we request a new cert for each subdomain during initialization. The prototype of the haproxy-route state component is also changed slightly to accomodate this ( This is because we need to parse the haproxy-route relation data during charm init and during the initialization of the certificates requirer ). Some minor bug fixes for certificates handling and some updates to unit test is also present.

### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -331,11 +331,6 @@ class HAProxyCharm(ops.CharmBase):
                 logger.exception(
                     "haproxy-route information not ready, skipping certificate request."
                 )
-                return [
-                    CertificateRequestAttributes(
-                        common_name=external_hostname, sans_dns=frozenset([external_hostname])
-                    )
-                ]
         return [
             CertificateRequestAttributes(
                 common_name=external_hostname, sans_dns=frozenset([external_hostname])
--- a/src/charm.py
+++ b/src/charm.py
@@ -25,7 +25,7 @@ from charms.traefik_k8s.v2.ingress import (
     IngressPerAppDataRemovedEvent,
     IngressPerAppProvider,
 )
-from interface_hacluster.ops_ha_interface import HAServiceReadyEvent, HAServiceRequires
+from interface_hacluster.ops_ha_interface import HAServiceRequires
 from ops.charm import ActionEvent
 from ops.model import Port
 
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -295,6 +308,34 @@ def _get_certificate_requests(self) -> typing.List[CertificateRequestAttributes]
         external_hostname = typing.cast(str, self.config.get("external-hostname", None))
         if not external_hostname:
             return []
+
+        if self.haproxy_route_provider.relations and not (
```
arturo-seijas, line 312: Why not just `if self.haproxy_route_provider.relations`?

src/charm.py
```diff
@@ -295,6 +308,34 @@ def _get_certificate_requests(self) -> typing.List[CertificateRequestAttributes]
...
+                        common_name=hostname_acl, sans_dns=frozenset([hostname_acl])
+                    )
+                    for backend in haproxy_route_requirer_information.backends
+                    for hostname_acl in backend.hostname_acls
+                ]
+            except (HaproxyRouteIntegrationDataValidationError, TLSNotReadyError):
+                logger.exception(
+                    "haproxy-route information not ready, skipping certificate request."
+                )
+                return [
```
arturo-seijas: You're already returning the same value in line 339, so this additional doesn't have any effect

src/tls_relation.py
```diff
@@ -53,16 +53,13 @@ def get_provider_cert_with_hostname(
...
-        if not provider_certificate:
-            return None
-        if not provider_certificate.certificate:
-            return None
-        if provider_certificate.certificate.common_name != hostname:
-            return None
-        return provider_certificate
+
+        provider_certificates, _ = self.certificates.get_assigned_certificates()
+        for provider_cert in provider_certificates:
```
arturo-seijas, line 58: This block looks like a good candidate for a comprehension list

src/charm.py
Thanhphan1147: Because if we're in an invalid state ( both `haproxy-route` and `ingress` are related for example ) then the charm will block but as this method is called during init, I'd rather skip parsing the relation data here if we know that the charm will block afterwards

src/tls_relation.py
Thanhphan1147: I'm not sure how a list comprehension would work here, as we need to return the first matching cert in the middle of the loop, can you elaborate a bit further?

src/charm.py
Thanhphan1147: Makes sense, I removed it.

src/charm.py
arturo-seijas: So it is just to avoid parsing data? I'd parse it regardless instead of adding complexity

src/tls_relation.py
arturo-seijas: `[provider_cert for provider_cert in provider_certificates if provider_cert.certificate and provider_cert.certificate.common_name == hostname]` or something like that?

Thanhphan1147: Hmm, I'm not sure about this since this means we'll go through the full list of certificates before return, while currently we return as soon as we get a match. What do you think?

src/charm.py
Thanhphan1147: Also we'd need to parse data and request for certificates under ingress mode as well, and if both relations are established we are in an ambiguous state since we don't know that we should request cert for ingress or for haproxy, so I think adding explicit conditions like this is necessary. Wdyt?

PR #150, opened 2025-07-08
ISD-3795 update certificate handling, update tests, request cert for haproxy-route backends using subdomains

Thanhphan1147:
Currently haproxy only request cert for the external-hostname config, if a haproxy-route requirer charm requested a subdomain the resulting fqdn should also have a certificate.

In this PR we request a new cert for each subdomain during initialization. The prototype of the haproxy-route state component is also changed slightly to accomodate this ( This is because we need to parse the haproxy-route relation data during charm init and during the initialization of the certificates requirer ). Some minor bug fixes for certificates handling and some updates to unit test is also present.

### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -331,11 +331,6 @@ class HAProxyCharm(ops.CharmBase):
                 logger.exception(
                     "haproxy-route information not ready, skipping certificate request."
                 )
-                return [
-                    CertificateRequestAttributes(
-                        common_name=external_hostname, sans_dns=frozenset([external_hostname])
-                    )
-                ]
         return [
             CertificateRequestAttributes(
                 common_name=external_hostname, sans_dns=frozenset([external_hostname])
--- a/src/charm.py
+++ b/src/charm.py
@@ -25,7 +25,7 @@ from charms.traefik_k8s.v2.ingress import (
     IngressPerAppDataRemovedEvent,
     IngressPerAppProvider,
 )
-from interface_hacluster.ops_ha_interface import HAServiceReadyEvent, HAServiceRequires
+from interface_hacluster.ops_ha_interface import HAServiceRequires
 from ops.charm import ActionEvent
 from ops.model import Port
 
@@ -39,9 +39,13 @@ from http_interface import (
 from state.config import CharmConfig
 from state.exception import CharmStateValidationBaseError
 from state.ha import HACLUSTER_INTEGRATION, HAPROXY_PEER_INTEGRATION, HAInformation
-from state.haproxy_route import HAPROXY_ROUTE_RELATION, HaproxyRouteRequirersInformation
+from state.haproxy_route import (
+    HAPROXY_ROUTE_RELATION,
+    HaproxyRouteIntegrationDataValidationError,
+    HaproxyRouteRequirersInformation,
+)
 from state.ingress import IngressRequirersInformation
-from state.tls import TLSInformation
+from state.tls import TLSInformation, TLSNotReadyError
 from state.validation import validate_config_and_tls
 from tls_relation import TLSRelationService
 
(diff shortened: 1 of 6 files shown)


seb4stien (approved):
Thanks for the live review.
As discussed it would be nice to have 2 comments to give the "why" you took this approach in the init.
And an additional integration test with the happy path would be nice to ensure we get the multiple certificates.

src/charm.py
```diff
@@ -295,6 +308,29 @@ def _get_certificate_requests(self) -> typing.List[CertificateRequestAttributes]
         external_hostname = typing.cast(str, self.config.get("external-hostname", None))
         if not external_hostname:
             return []
+
+        if self.haproxy_route_provider.relations and not (
```
arturo-seijas: Following up on https://github.com/canonical/haproxy-operator/pull/144/files#r2179342543, you are talking about mode, but that hasn't been implemented yet. Maybe I'm failing to see why this is needed now

src/tls_relation.py
```diff
@@ -53,16 +53,13 @@ def get_provider_cert_with_hostname(
...
-        if not provider_certificate:
-            return None
-        if not provider_certificate.certificate:
-            return None
-        if provider_certificate.certificate.common_name != hostname:
-            return None
-        return provider_certificate
+
+        provider_certificates, _ = self.certificates.get_assigned_certificates()
+        for provider_cert in provider_certificates:
```
arturo-seijas: Following up on https://github.com/canonical/haproxy-operator/pull/144/files#r2179358147,  I'm ok with it as it is, but it's not like there are lots of certificates. We should aim to have code as simple as possible given that those optimizations won't have any impact on the overall charm performances

src/charm.py
Thanhphan1147: An example would be if both the 'ingress' relation and the 'haproxy-route' relation is established. For ingress if we're under subdomain routing we'd want to request certificate for 'model-app.<external-hostname>' but for 'haproxy-route' we need to request certificate for '<subdomain>.<external-hostname>'.

So my thought process is that if we're in that case where it's ambiguous then we don't know what certificate to request so it's better to have explicit condition here.

Potentially I can detect since the beginning of the method if we are in an ambiguos state and return an empty list, and keep the condition simple afterwards

Thanhphan1147: I updated the code, let me know what you think.

src/charm.py
```diff
@@ -295,6 +310,42 @@ def _get_certificate_requests(self) -> typing.List[CertificateRequestAttributes]
...
+        # This is the same logic as in self._validate_state, but we don't reuse that method
+        # as we want to avoid setting charm status during initialization
+        if (
+            bool(self.haproxy_route_provider.relations)
+            + bool(self.reverseproxy_requirer.relations)
+            + bool(self._ingress_provider.relations)
+            > 1
+        ):
+            # The charm is in an invalid state, we don't request for certificates here
+            return []
```
swetha1654: You could perhaps remove setting the blocked charm status in the `_validate_state` function and raise an Exception instead, so that it is reusable. You set the blocked status inside the `validate_config_and_tls` instead.

swetha1654 (approved):

src/tls_relation.py
Thanhphan1147: I updated the code to use a list comprehension instead, let me know what you think!

src/charm.py
Thanhphan1147: That's probably the right thing to do :sweat_smile: I refactored the code to reflect that. Thanks!

Thanhphan1147:
I'm closing the comments now and merging the PR, please let me know if you want to continue the discussion internally @arturo-seijas / @swetha1654!

swetha1654:
@Thanhphan1147 I have one more comment - `HaproxyTooManyIntegrationsError` needs to be handled in the `validate_config_and_tls` function.

Thanhphan1147:
@swetha1654 that wouldn't be a problem i think since the exception subclasses 'CharmStateValidationBaseError' which is handled in the decorator.

swetha1654:
@Thanhphan1147 right didn't see that, thanks!

PR #152, opened 2025-07-11
ISD-3826 Update lib to add hostname and additional hostnames attribute, update haproxy logic to handle new attributes

Thanhphan1147:
Applicable spec: ISD229 ( internal )

replace the "subdomain" attribute with the "hostname" and "additional_hostnames" attribute. Haproxy now sends an empty list as the `proxied_endpoints` when the requirer data is ignored ( due to validation error or missing hostname )

Updated tests.

### Checklist


arturo-seijas (approved):
Can you update the relation interfaces definition too?

lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -170,6 +171,9 @@ def value_contains_invalid_characters(value: Optional[str]) -> Optional[str]:
     return value
 
 
+VALIDSTR = Annotated[str, BeforeValidator(value_contains_invalid_characters)]
```
seb4stien, line 174: In terms of security it's better to explicitely allow some characters. We had a similar discussion for the configurator, you can probably reuse some of the validators here http://github.com/canonical/ingress-configurator-operator/blob/main/src/validators.py#L14

We should also probably have specific validators for the different kind of strings we may have (I'll drop comments below).

lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -304,9 +308,7 @@ class ServerHealthCheck(BaseModel):
     fall: int = Field(
         description="How many failed health checks before server is considered down.", default=3
     )
-    path: Annotated[Optional[str], BeforeValidator(value_contains_invalid_characters)] = Field(
-        description="The health check path.", default=None
-    )
+    path: Optional[VALIDSTR] = Field(description="The health check path.", default=None)
```
seb4stien, line 311: Specific validator for path?

lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -368,7 +370,7 @@ class LoadBalancingConfiguration(BaseModel):
         description="Configure the load balancing algorithm for the service.",
         default=LoadBalancingAlgorithm.LEASTCONN,
     )
-    cookie: Annotated[Optional[str], BeforeValidator(value_contains_invalid_characters)] = Field(
+    cookie: Optional[VALIDSTR] = Field(
```
seb4stien, line 373: Specific validator for the cookie? (it's just an id here, right?)

lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -453,10 +455,8 @@ class RewriteConfiguration(BaseModel):
     method: HaproxyRewriteMethod = Field(
         description="Which rewrite method to apply.One of set-path, set-query, set-header."
     )
-    expression: Annotated[str, BeforeValidator(value_contains_invalid_characters)] = Field(
-        description="Regular expression to use with the rewrite method."
-    )
-    header: Annotated[Optional[str], BeforeValidator(value_contains_invalid_characters)] = Field(
+    expression: VALIDSTR = Field(description="Regular expression to use with the rewrite method.")
```
seb4stien, line 458: Specific validator for regex? (this one may be tricky :) ).

lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -453,10 +455,8 @@ class RewriteConfiguration(BaseModel):
...
+    header: Optional[VALIDSTR] = Field(
```
seb4stien, line 459: Specific validator for header?

lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -482,17 +484,18 @@ class RequirerApplicationData(_DatabagModel):
         server_maxconn: Optional maximum number of connections per server.
     """
 
-    service: Annotated[str, BeforeValidator(value_contains_invalid_characters)] = Field(
-        description="The name of the service."
-    )
+    service: VALIDSTR = Field(description="The name of the service.")
```
seb4stien, line 487: Specific validator for hostname?

lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -482,17 +484,18 @@ class RequirerApplicationData(_DatabagModel):
...
     ports: list[int] = Field(description="The list of ports listening for this service.")
     hosts: list[IPvAnyAddress] = Field(
         description="The list of backend server addresses. Currently only support IP addresses.",
         default=[],
     )
-    paths: list[str] = Field(description="The list of paths to route to this service.", default=[])
-    subdomains: list[str] = Field(
-        description="The list of subdomains to route to this service.", default=[]
+    paths: list[VALIDSTR] = Field(
```
seb4stien, line 493: Path validator?

lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -482,17 +484,18 @@ class RequirerApplicationData(_DatabagModel):
...
+        description="The list of paths to route to this service.", default=[]
+    )
+    hostname: Optional[VALIDSTR] = Field(description="Hostname of this service.", default=None)
```
seb4stien, line 496: Hostname validator?

lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -482,17 +484,18 @@ class RequirerApplicationData(_DatabagModel):
...
+    additional_hostnames: list[VALIDSTR] = Field(
```
seb4stien, line 497: Hostname validator?

lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -513,7 +516,7 @@ class RequirerApplicationData(_DatabagModel):
     retry: Optional[Retry] = Field(
         description="Configure retry for incoming requests.", default=None
     )
-    deny_paths: list[str] = Field(
+    deny_paths: list[VALIDSTR] = Field(
```
seb4stien, line 519: Path validator?

src/charm.py
```diff
@@ -235,8 +235,11 @@ def _reconcile(self) -> None:
                     self._ingress_provider
                 )
                 self.unit.set_ports(80, 443)
+                # In ingress mode it's expected for tls_information.hostnames
+                # to only contains the `external-hostname` charm config.
+                # Validation of tls_information will fail otherwise.
                 self.haproxy_service.reconcile_ingress(
-                    config, ingress_requirers_information, tls_information.external_hostname
+                    config, ingress_requirers_information, tls_information.hostnames[0]
```
seb4stien, line 242: Is the order guaranteed?

src/state/haproxy_route.py
```diff
@@ -205,6 +202,7 @@ def from_provider(
             stick_table_entries: list[str] = []
             requirers = haproxy_route.get_data(haproxy_route.relations)
             backends: list[HAProxyRouteBackend] = []
+            relation_ids_with_invalid_data = requirers.relation_ids_with_invalid_data
```
seb4stien, line 205: Not sure to get this one: will the relation_ids_with_invalid_data will be computed somewhere else?

src/state/tls.py
```diff
@@ -76,31 +83,40 @@ def from_charm(
...
             TLSNotReadyError: if the charm is not ready to handle TLS.
         """
-        tls_requirer_integration = charm.model.get_relation(TLS_CERTIFICATES_INTEGRATION)
-        external_hostname = typing.cast(str, charm.config.get("external-hostname", ""))
-
-        if not re.match(HOSTNAME_REGEX, external_hostname):
+        tls_requirer_integration = charm.model.get_relation(certificates.relationship_name)
+        if not certificates.certificate_requests:
+            logger.error("The charm did not request any certificates.")
+            raise TLSNotReadyError("The charm did not request any certificates.")
```
seb4stien, line 107: I may be lost in the PR: is it actually an error? Couldn't we have a ha-proxy with just http sites? Or are we already in a case here where we know we have some TLS but we are missing info (and the message may be slightly different in that case).

tests/integration/haproxy_route_requirer.py
```diff
@@ -37,7 +37,8 @@ def update_relation(self):
         self._haproxy_route.provide_haproxy_route_requirements(
             service="any",
             ports=[80],
-            subdomains=["ok", "ok2"],
+            hostname="ok.haproxy.internal",
+            additional_hostnames=["ok2.haproxy.internal"],
```
seb4stien: I would add more than one to ensure all of them are handled properly.

seb4stien (changes requested):

lib/charms/haproxy/v1/haproxy_route.py
Thanhphan1147: The validator is not related to the relation data attributes themselves and is mainly to avoid reserved characters in the haproxy config, so that the requirer can't set something like
```
service: "backend\nbackend injected\n\tserver <malicious IP address>"
```
for example. Other reserved character are also part of the banlist as they will break the haproxy config.

seb4stien: My doubt is: are we sure we are exhaustive in the banlist?

src/state/tls.py
Thanhphan1147: > Couldn't we have a ha-proxy with just http sites?

No, for ingress and haproxy-route we enfore TLS which means the charm should block if it didn't raise any cert request ( either `external-hostname` is not configured or requirer hasn't sent over any hostname ).

As we don't have any cert request it's up to the operator to either configure `external-hostname` or to relate haproxy to a requirer that requests a hostname ( that's why the status is blocked here rather than a waiting status since the responsibility falls on the operator to take action )

src/state/haproxy_route.py
Thanhphan1147: Yep, as the library should not know about what kind of charm config the provider using the library have. The library keeps track of the relation_ids where the data is invalid, and the haproxy charm on top of that keeps track of relation_ids where the requirer didn't set `hostname` and `external-hostname` is not set.

src/charm.py
Thanhphan1147: not only be the order that should be guaranteed, but also the list is expected to have exactly 1 element.

lib/charms/haproxy/v1/haproxy_route.py
Thanhphan1147: I got the banlist from the list of characters that need to be escaped in the haproxy documentation : https://www.haproxy.com/documentation/haproxy-configuration-manual/latest/#2.2

Thanhphan1147: so basically `<space> , ' " # \ \n \r \t` and I skipped ascii hex code for now as it's too complicated to implement.

Thanhphan1147: Now that I think about it, as this is specific to haproxy, we can also move the validation to the haproxy charm, and leave room for more meaningful validation in the lib. WDYT?

lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -34,7 +34,8 @@ def __init__(self, *args):
         address=<required>,
         port=<required>,
         paths=<optional>,
-        subdomains=<optional>,
```
swetha1654, line 37: since we're removing subdomains, we would have to take care of this change in the ingress-configurator as well right? Could you create a JIRA for the same?

src/charm.py
seb4stien: Shouldn't we get one for each additional_domain?

lib/charms/haproxy/v1/haproxy_route.py
seb4stien: I guess it depends on where we want to raise the potential errors. If we keep it in the library, could it raise error on the requirer side? I feel that would be better in terms of user experience that having the validation (silently) failing on the provider side.

Thanhphan1147: Yes having them here will make it so that `DataValidationError` is raised when calling `provide_haproxy_route_requirements` with invalid haproxy characters

Thanhphan1147: in that case I'd keep it like this.

lib/charms/haproxy/v1/haproxy_route.py
Thanhphan1147: closing this as per our discussion above.

lib/charms/haproxy/v1/haproxy_route.py
Thanhphan1147: closing this as per our discussion above.

lib/charms/haproxy/v1/haproxy_route.py
Thanhphan1147: closing this as per our discussion above.

lib/charms/haproxy/v1/haproxy_route.py
Thanhphan1147: closing this as per our discussion above.

lib/charms/haproxy/v1/haproxy_route.py
Thanhphan1147: closing this as per our discussion above.

src/charm.py
Thanhphan1147: here we are in ingress mode, so there's no notion of `additional_hostnames` here as that only exists in haproxy-route mode. And we made them mutually exclusive

lib/charms/haproxy/v1/haproxy_route.py
Thanhphan1147: Yes, I created ISD-3855 for that

swetha1654 (approved):

tests/integration/haproxy_route_requirer.py
Thanhphan1147: updated!

Thanhphan1147:
Hi @seb4stien, can you take a look at the PR again after the changes? thanks!

seb4stien (approved):

lib/charms/haproxy/v1/haproxy_route.py
Thanhphan1147: closing this as per our discussion above.

lib/charms/haproxy/v1/haproxy_route.py
Thanhphan1147: closing this as per our discussion above.

lib/charms/haproxy/v1/haproxy_route.py
Thanhphan1147: closing this as per our discussion above.

lib/charms/haproxy/v1/haproxy_route.py
Thanhphan1147: closing this as per our discussion above.

Issue #88, opened 2025-03-17
Add the ingress-per-unit integration


saltiyazan:
### Enhancement Proposal

Hello Team!

I would like to request the implementation of the `ingress-per-unit` integration as we will need it to support reaching each unit separately for a Vault deployed behind an ingress.

Is this an item on your roadmap for the current or the next sprint?

Thanks!

yanksyoon:
@seb4stien Hi Sebastien! Would you be able to help answer the question regarding the roadmap please? Thank you! :)

seb4stien:
Hi. We included it in the prioritization process for the next cycle.

swetha1654:
@saltiyazan this is being tackled in this pulse, stay tuned for more updates!

swetha1654:
@saltiyazan the ingress per unit relation has been implemented. You should be able to use it with the 199 revision in the 2.8/edge channel of the HAProxy charm. Feel free to reopen this issue if you face any issues with the relation.

PR #153, opened 2025-07-15
Ingress per unit relation implementation

swetha1654:
Applicable spec: <link>

### Overview

Addresses #88 

- Add ingress per unit relation.
- Remove `src-docs` references.

### Checklist

- [ ] The documentation is generated using `src-docs`


src/charm.py
```diff
@@ -226,24 +232,34 @@ def _reconcile(self) -> None:
         self._reconcile_ha(ha_information)
 
         config = CharmConfig.from_charm(self)
+
+        if self.model.get_relation(TLS_CERT_RELATION):
+            # Reconcile certificates in case the certificates relation is present
```
Thanhphan1147: The reason why this code is not here but in the switch cases is because we also support the legacy reverseproxy relation, and we don't want to enfore tls for that specific mode.

src/charm.py
```diff
@@ -226,24 +232,34 @@ def _reconcile(self) -> None:
...
+                    config,
+                    ingress_per_unit_requirers_information,
+                    tls_information.external_hostname,
+                )
             case ProxyMode.LEGACY:
-                if self.model.get_relation(TLS_CERT_RELATION):
-                    # Reconcile certificates in case the certificates relation is present
-                    tls_information = TLSInformation.from_charm(self, self.certificates)
-                    self._tls.certificate_available(tls_information)
-
```
swetha1654: @Thanhphan1147 i see TLS is enforced in LEGACY mode as well here.

src/charm.py
swetha1654: See: https://github.com/canonical/haproxy-operator/pull/153#discussion_r2212359432

src/charm.py
```diff
@@ -359,14 +369,33 @@ def _on_ingress_data_provided(self, event: IngressPerAppDataProvidedEvent) -> No
...
+                                unit.name,
+                            )
+                            continue
+                        integration_data = self._ingress_per_unit_provider.get_data(relation, unit)
+                        path_prefix = f"{integration_data['model']}-{integration_data['name']}"
+                        self._ingress_per_unit_provider.publish_url(
+                            relation,
+                            integration_data["name"],
+                            f"https://{tls_information.external_hostname}/{path_prefix}",
+                        )
```
seb4stien: Open quesiton: should we have a final else to raise some error or at least to log something if we are not in any of the previous cases?

src/charm.py
```diff
@@ -381,10 +410,17 @@ def _validate_state(self) -> ProxyMode:
...
-        if is_ingress_related + is_legacy_related + is_haproxy_route_related > 1:
+        if (
+            is_ingress_per_unit_related
+            + is_ingress_related
+            + is_legacy_related
+            + is_haproxy_route_related
+            > 1
+        ):
             msg = (
                 "Only one integration out of 'ingress', 'reverseproxy' or 'haproxy-route' "
```
seb4stien: You miss one relation type.

templates/haproxy_ingress_per_unit.cfg.j2
```diff
@@ -0,0 +1,31 @@
...
+    use_backend {{ backend.backend_name }} if path_{{ backend.backend_name }} host_{{ config_external_hostname }}
+    default_backend default
+{% endfor %}
+
+{% for backend in ingress_per_unit_requirers_information.backends %}
+backend {{ backend.backend_name }}
+{% if backend.strip_prefix %}
+    http-request replace-path /{{ backend.backend_path }}(/)?(.*) /\2
+{% endif %}
+    server unit {{ backend.hostname_or_ip }}:{{ backend.port }} check
```
Thanhphan1147, line 29: We probably need the unit index to make the server name unique

src/charm.py
seb4stien: Isn't ingress_per_unit compatible with ingress?
Would it make sense to support both on the same ha-proxy instance?
cc @Thanhphan1147

tests/integration/ingress_per_unit_requirer.py
Thanhphan1147, line 1: Since we want to move to jubilant at some point, I think we should start  using jubilant for new integration tests.

tests/integration/test_ingress_per_unit.py
```diff
@@ -0,0 +1,65 @@
...
+        assert parsed_url.scheme == "https"
+
+        path_suffix = f"{parsed_url.path}/ok"
+
+        if isinstance(unit_ip, ipaddress.IPv6Address):
+            backend_url = f"http://[{unit_ip}]{path_suffix}"
+        else:
+            backend_url = f"http://{unit_ip}{path_suffix}"
+
+        print(backend_url)
```
seb4stien: Leftover? You may want to log, but probably not to print.

seb4stien (approved):

src/charm.py
Thanhphan1147: Hmm, indeed. Iirc there was a reason why this block was moved from here to the switch cases. Let me take a look at the commit history

tests/integration/test_ingress_per_unit.py
swetha1654: oops, thanks for pointing it out

src/charm.py
swetha1654: can you give me an example when we would require a final else? I believe this custom event will only be called when ingress and ingress per unit have their data ready.

src/charm.py
swetha1654: added the relation type. I'm not sure why an application would need an ingress per app as well when it has an ingress per unit.

templates/haproxy_ingress_per_unit.cfg.j2
swetha1654: We are defining a backend per unit, and I believe the server names are scoped within the backend. So i don't think its necessary for the server names to be unique. WDYT?

tests/integration/ingress_per_unit_requirer.py
swetha1654: I am currently using some helper functions which are also used by other integration tests. Migrating this test to jubilant would involve migrating some of the others as well. Or we would have to duplicate some code to do the same thing in jubilant and ope_test. Shall I address this in a seperate PR?

src/charm.py
Thanhphan1147: Ah I know why, you might notice both in default mode and in legacy mode `tls_information` is [only instanciated when the relation is there](https://github.com/canonical/haproxy-operator/pull/153#discussion_r2212359432) ( as we don't enforce TLS but if they are there we'll write them to the unit's fs ). While under `ingress` and `haproxy-route` mode we instanciate them regardless to enforce TLS. 

If we conditionally instanciate `tls_information` before the `switch` we'll have to add something like `assert tls_information` at the beginnig of the switch case to ensure that it exists, or do some other kind of verification logic to ensure that the state is properly instanciated which seems counter-intuitive.

src/charm.py
Thanhphan1147: answered in above comment.

src/charm.py
swetha1654: Understood, will revert it back

templates/haproxy_ingress_per_unit.cfg.j2
Thanhphan1147: Ah you're right, let's keep it like this.

tests/integration/ingress_per_unit_requirer.py
Thanhphan1147: personally I'd add the test with the necessary fixtures using jubilant and it's fine if the pytest fixtures and the jubilant fixtures overlap. Otherwise we'll keep adding tests based on the existing fixtures since it's convenience to do and we'd just pile up our technical debt. But I'd like to hear @seb4stien and @arturo-seijas opinion on this as well. 

What I've been doing for a while now is that all new unit tests in this repo were made with scenario and they co-exist with the old harness tests.

src/charm.py
Thanhphan1147: I think it make sense to isolate them, I agree that I don't really see the use-case where you'd need both at the same time

.github/workflows/integration_test.yaml
```diff
@@ -12,4 +12,4 @@ jobs:
       juju-channel: 3/stable
       self-hosted-runner: true
       charmcraft-channel: latest/edge
-      modules: '["test_action.py","test_charm.py","test_config.py","test_cos.py","test_ha.py","test_haproxy_route.py","test_http_interface.py","test_ingress.py"]'
+      modules: '["test_action.py","test_charm.py","test_config.py","test_cos.py","test_ha.py","test_haproxy_route.py","test_http_interface.py","test_ingress.py", "test_ingress_per_unit.py"]'
```
arturo-seijas, line 15: Can you order these alphabetically? test_http_interface is out of place :)

src/haproxy.py
```diff
@@ -107,6 +109,30 @@ def reconcile_legacy(self, config: CharmConfig, services: list) -> None:
         self._validate_haproxy_config()
         self._reload_haproxy_service()
 
+    def reconcile_ingress_per_unit(
```
arturo-seijas: This method is almost the same as  `reconcile_ingress`, only the template changes. You can refactor it and accept also the template as an argument so that both can be merged. This will probably enable some more refactoring on the caller method.

src/charm.py
```diff
@@ -350,7 +367,9 @@ def _get_certificate_requests(self) -> typing.List[CertificateRequestAttributes]
         ]
 
     @validate_config_and_tls(defer=True)
-    def _on_ingress_data_provided(self, event: IngressPerAppDataProvidedEvent) -> None:
+    def _on_ingress_data_provided(
```
arturo-seijas: There are two very distinct executions paths here. Is there any reason why they are not two different methods?

src/charm.py
```diff
@@ -381,20 +419,30 @@ def _validate_state(self) -> ProxyMode:
             ProxyMode: The resulting proxy mode.
         """
         is_ingress_related = bool(self._ingress_provider.relations)
+        is_ingress_per_unit_related = bool(self._ingress_per_unit_provider.relations)
         is_legacy_related = bool(self.reverseproxy_requirer.relations)
         is_haproxy_route_related = bool(self.haproxy_route_provider.relations)
 
-        if is_ingress_related + is_legacy_related + is_haproxy_route_related > 1:
+        if (
```
arturo-seijas, line 461: This kind of validation should live in the state, actually, the whole method

src/charm.py
```diff
@@ -381,20 +419,30 @@ def _validate_state(self) -> ProxyMode:
...
+                "Only one integration out of 'ingress', 'ingress-per-unit', "
+                "'reverseproxy' or 'haproxy-route' can be active at a time."
             )
             logger.error(msg)
             raise HaproxyTooManyIntegrationsError(msg)
 
         if is_ingress_related:
             return ProxyMode.INGRESS
 
+        if is_ingress_per_unit_related:
```
arturo-seijas, line 478: The mode should probably be a property of the state. Returning the mode as part of the validation doesn't seem very obvious to me

tests/integration/test_ingress_per_unit.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Assert that the requirer endpoints are available.
+    """
+    application = configured_application_with_tls
+    requirer_app = any_charm_ingress_per_unit_requirer
+    await configured_application_with_tls.model.add_relation(
+        f"{configured_application_with_tls.name}:ingress-per-unit",
+        f"{requirer_app.name}:require-ingress-per-unit",
+    )
+    await configured_application_with_tls.model.wait_for_idle(
```
arturo-seijas, line 33: It'll be safer to wait also for `requirer_app`

tests/unit/test_ingress_per_unit_scenario.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Unit tests for the ingress-per-unit mode of haproxy charm."""
```
arturo-seijas, line 4: Can you rename this test file to `test_charm.py`. This actually tests the charm.py file, so let's follow conventions. Also, there's no need to suffix the tests with `_scenario`.
Also, you should unit tests the state/ingress_peR_unit.py file by itself, without doing it though the charm,, as it exposes a public interface.

tests/unit/test_ingress_scenario.py
```diff
@@ -1,7 +1,7 @@
 # Copyright 2025 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-"""Unit tests for the haproxy charm."""
+"""Unit tests for the ingress mode of haproxy charm."""
```
arturo-seijas, line 4: Same as above, can you rename the file?

.github/workflows/integration_test.yaml
swetha1654: Am i missing something? It looks alphabetically sorted to me.

src/charm.py
swetha1654: actually, there isn't. Let me split it into 2 methods.

src/charm.py
swetha1654: @arturo-seijas  I don't think we have a state for HAproxy charm as a whole, we have a seperate state for each mode like we did initially with the ingress configurator charm and that mode is decided by this method. I think this is another discussion we need to have in the planning meeting.

src/charm.py
swetha1654: Same as above, there is no seperate state for the charm.

tests/unit/test_ingress_per_unit_scenario.py
swetha1654: I think @Thanhphan1147 started naming the files with `_scenario` suffix to distinguish the tests using the new and old framework. And there is already a `test_charm_scenario.py` that already exists. Apart from that, the convention in this repo seems to be one file for every MODE. Changing this to `test_charm.py` would mean moving the other tests to this file as well. WDYT @Thanhphan1147 ?

seb4stien (approved):

src/charm.py
```diff
@@ -213,7 +229,7 @@ def _on_http_backend_removed(self, _: HTTPBackendRemovedEvent) -> None:
         """Handle data_removed event for reverseproxy integration."""
         self._reconcile()
 
-    def _reconcile(self) -> None:
+    def _reconcile(self) -> None:  # pylint: disable=too-many-locals, too-many-statements
```
Thanhphan1147, line 232: Note: not blocking the merge but these comments probably means that we need to start moving the logic in each switch case into its own method

swetha1654: makes sense. I will handle this as well in my refactoring PR.

src/haproxy.py
```diff
@@ -110,23 +116,39 @@ def reconcile_legacy(self, config: CharmConfig, services: list) -> None:
...
+                type.
         """
         template_context = {
             "config_global_max_connection": config.global_max_connection,
             "ingress_requirers_information": ingress_requirers_information,
             "config_external_hostname": external_hostname,
             "haproxy_crt_dir": HAPROXY_CERTS_DIR,
         }
-        self._render_haproxy_config(HAPROXY_INGRESS_CONFIG_TEMPLATE, template_context)
+        if isinstance(ingress_requirers_information, IngressRequirersInformation):
```
Thanhphan1147, line 142: Would something like this better?
```
if not isinstance(...) and not isinstance(...):
              raise InvalidRequirerInformationError(
                "Expected ingress requirer or ingress per unit requirer "
                f"but received {ingress_requirers_information} instead."
            )

template = HAPROXY_INGRESS_CONFIG_TEMPLATE if isinstance(ingress_requirers_information, IngressPerUnitRequirersInformation) else HAPROXY_INGRESS_PER_UNIT_CONFIG_TEMPLATE
self._render_haproxy_config(template, template_context)
```

Though I'm not even sure about needing the `if` here because the type is already defined in the method prototype and python will give RuntimeError if called with the wrong parameter type right?

Thanhphan1147: Nevermind about the type, seems like you can pass anything and python won't complain. Now I wonder if checking for it makes sense because that logic is happening internally to the charm which means we have full control over what gets passed as parameters for this method

src/state/ingress_per_unit.py
```diff
@@ -0,0 +1,90 @@
...
+        backends = []
+        for integration in ingress_per_unit_provider.relations:
+            for unit in integration.units:
+                try:
+                    integration_data: RequirerData = ingress_per_unit_provider.get_data(
+                        integration, unit
+                    )
+                    unit_name = "_".join(integration_data["name"].split("/"))
+                    backend_name = f"{integration_data['model']}_{unit_name}"
+                    backend_path = f"{integration_data['model']}-{integration_data['name']}"
```
Thanhphan1147, line 76: backend_path needs the unit index here right? Or am I missing something?

swetha1654: `integration_data['name']` is the unit name (eg - "requirer/0"). So the index will be present.

src/haproxy.py
swetha1654: yeah i think it makes sense to not have the if condition at all. Should i just add this?
```
template = HAPROXY_INGRESS_CONFIG_TEMPLATE if isinstance(ingress_requirers_information, IngressPerUnitRequirersInformation) else HAPROXY_INGRESS_PER_UNIT_CONFIG_TEMPLATE
self._render_haproxy_config(template, template_context)
```

.github/workflows/integration_test.yaml
arturo-seijas: I don't know what _I was thinking about

arturo-seijas (approved):

src/haproxy.py
Thanhphan1147: yes, looks good to me.

src/state/ingress_per_unit.py
Thanhphan1147: Ah you're right.

Thanhphan1147 (approved):

PR #158, opened 2025-07-21
Refactoring HAProxy source code and tests

swetha1654:
Applicable spec: <link>

### Overview

Mainly addressing the changes from the ingress-per-unit [PR](https://github.com/canonical/haproxy-operator/pull/153) that were out of the scope of that PR.

- Migrate ingress-per-unit integration tests to jubilant. 
- Migrated the existing tests using ops_test framework to the `/legacy` directory.
- Move all unit test modules for individual modes into `test_charm_scenario.py` and `test_state_scenario.py`.
- Have a charmState and perform mode validation in the state rather than a method in the charm class.
- Break down the reconcile function into smaller modules.

### Checklist


src/charm.py
```diff
@@ -241,125 +223,135 @@ def _reconcile(self) -> None:  # pylint: disable=too-many-locals, too-many-state
...
+        )
+        self.unit.set_ports(80, 443)
+        # In ingress mode it's expected for tls_information.hostnames
+        # to only contains the `external-hostname` charm config.
+        # Validation of tls_information will fail otherwise.
+        self.haproxy_service.reconcile_ingress(
+            charm_state, ingress_requirers_information, tls_information.hostnames[0]
+        )
+
+    def _configure_ingress_per_unit(self, charm_state: CharmState) -> None:
```
arturo-seijas: You could merge this `_configure_ingress`. the only different I see is the attribute from which the `IngressRequirersInformation` is instantiated, and that can be parametrized

src/state/charm_state.py
```diff
@@ -0,0 +1,211 @@
...
+        Args:
+            global_max_connection: The config to validate.
+
+        Raises:
+            ValueError: When the configured value is not between 0 and "fs.file-max".
+
+        Returns:
+            int: The validated global_max_connection config.
+        """
+        if global_max_connection < 0:
```
arturo-seijas: You can remove this and set a field constraint on the attribute definition, something like
global_max_connection: in = Field(gt=0)

src/state/ingress_per_unit.py
```diff
@@ -87,4 +92,9 @@ def from_provider(
                     raise IngressPerUnitIntegrationDataValidationError(
                         "Validation of ingress per unit relation data failed."
                     ) from exc
+                except ValueError as exc:
+                    logger.error(str(exc))
+                    raise IngressPerUnitIntegrationDataValidationError(
+                        "App port from the ingress per unit relation is invalid."
```
arturo-seijas: We'll probably face issues later on with this message not being updated if additional ValueErrors are possible. I'd set a more generic message

tests/integration/helper.py
```diff
@@ -72,45 +73,64 @@ def send(
...
+        list: The parsed ingress URLs per unit.
     """
-    unit_name = ingress_requirer_application.units[0].name
-    _, stdout, _ = await ops_test.juju("show-unit", unit_name, "--format", "json")
-    unit_information = json.loads(stdout)[unit_name]
-    ingress_integration_data = json.loads(
-        unit_information["relation-info"][0]["application-data"]["ingress"]
-    )
-    return urlparse(ingress_integration_data["url"])
+    unit_name = app_name + "/0"
```
arturo-seijas: f-string?

tests/integration/legacy/conftest.py
```diff
@@ -0,0 +1,433 @@
...
+) -> typing.AsyncGenerator[Application, None]:
+    """Deploy the charm."""
+    app_name = "haproxy"
+    if pytestconfig.getoption("--no-deploy") and app_name in model.applications:
+        logger.warning("Using existing application: %s", app_name)
+        yield model.applications[app_name]
+        return
+    # Deploy the charm and wait for active/idle status
+    application = await model.deploy(f"./{charm}", trust=True)
+    await model.wait_for_idle(
```
arturo-seijas: The deployment will be faster is you don't wait until after the integration is established

tests/integration/legacy/conftest.py
```diff
@@ -0,0 +1,433 @@
...
+    pytestconfig: pytest.Config,
+    model: Model,
+) -> Application:
+    """Deploy self-signed-certificates."""
+    app_name = "self-signed-certificates"
+    if pytestconfig.getoption("--no-deploy") and app_name in model.applications:
+        logger.warning("Using existing application: %s", app_name)
+        return model.applications[app_name]
+    application = await model.deploy("self-signed-certificates", channel="1/edge")
+    await model.wait_for_idle(apps=[application.name], status="active")
```
arturo-seijas: Same here

tests/unit/test_state_scenario.py
```diff
@@ -16,13 +16,17 @@
     HaproxyRouteIntegrationDataValidationError,
     HaproxyRouteRequirersInformation,
 )
+from state.ingress_per_unit import (
```
arturo-seijas: Shouldn't this be the test_charm.py file?

seb4stien (approved):

tests/unit/test_state_scenario.py
swetha1654: no i believe this is correct. The `test_charm.py` contains all the unit tests validating the code in the `charm.py` file. The `test_state.py` validates all the files in the `state/` directory.

tests/integration/legacy/conftest.py
swetha1654: turns out this wait_for_idle is required since some of the tests access the "application" fixture directly to perform some validations. Keeping this `wait_for_idle`

tests/unit/test_state_scenario.py
arturo-seijas: We shouldn't need to rely on scenario nor harness for these tests. The way I see it, what is been tested is the charm as a whole because we interacting with the charm interfaces (adding relations, mocking application data...), the state doesn't have any of that. Also, the file is named scenario but this are harness tests

swetha1654: The tests in test_state are not testing the charm as a whole. They are simply initialising the required relations and units to then test the if the State classes for every mode are returning the expected values (IngressPerUnitRequirersInformation / HaproxyRouteRequirersInformation etc). 
Whereas the tests in `test_charm.py` are mocking hooks and triggering them to then evaluate if the charm has the right status / calls the reconcile function.

swetha1654: @arturo-seijas wrt the second point indeed you are right, it uses harness, so the name is incorrect. I will rename it back to test_state.

src/state/ingress_per_unit.py
```diff
@@ -87,4 +92,7 @@ def from_provider(
                     raise IngressPerUnitIntegrationDataValidationError(
                         "Validation of ingress per unit relation data failed."
                     ) from exc
+                except ValueError as exc:
+                    logger.error(str(exc))
+                    raise IngressPerUnitIntegrationDataValidationError(str(exc)) from exc
```
Thanhphan1147: I think `from exc` should already include the exception message and `logger.error` should be used to provide more context and we should remove it if we don't have much to add.

src/state/validation.py
```diff
@@ -96,6 +96,13 @@ def wrapper(instance: C, *args: typing.Any) -> None:
                     str(exc),
                 )
                 return None
+            except ValueError as exc:
+                if defer:
+                    event, *_ = args
+                    event.defer()
+                instance.unit.status = ops.BlockedStatus(str(exc))
+                logger.exception("ValueError occurred: %s", str(exc))
+                return None
```
Thanhphan1147: a base exception like `ValueError` should not flow all the way here and should be wrapped by another Exception

src/state/ingress_per_unit.py
```diff
@@ -87,4 +89,6 @@ def from_provider(
                     raise IngressPerUnitIntegrationDataValidationError(
                         "Validation of ingress per unit relation data failed."
                     ) from exc
+                except ValueError as exc:
+                    raise IngressPerUnitIntegrationDataValidationError(str(exc)) from exc
```
Thanhphan1147: Can you let me know what part of the code is supposed to raise ValueError? iirc pydantic should always wrap it with `ValidationError`

tests/unit/test_state_scenario.py
```diff
@@ -163,3 +168,107 @@ def test_haproxy_route_from_provider_duplicate_backend_names(
             external_hostname=MOCK_EXTERNAL_HOSTNAME,
             peers=[],
         )
+
+
+def test_ingress_per_unit_from_provider(harness: Harness):
```
Thanhphan1147: shouldn't new unit tests be done with scenario?

Thanhphan1147 (approved):

tests/unit/test_state_scenario.py
arturo-seijas: Indeed. I got confused there. Neither harness nor scenario should be used to test anything other than the charm itself, otherwise that'd introduce an unnecessary dependency. In cases like this, the charm should be just mocked, like in [this example](https://github.com/canonical/ingress-configurator-operator/blob/main/tests/unit/test_state.py#L19)

swetha1654: gotcha thanks! will do the same here :)

swetha1654: updated the tests suite. Moved the harness tests to legacy folder and used mocks for the `test_state.py` file.

tests/unit/test_state_scenario.py
swetha1654: Yup arturo gave me the same comment. I totally missed that. I have updated the tests now to use Mocks instead of harness. I also moved the harness tests to a `legacy/` folder and renamed the `test_*_scenario` files to `test_*` to be in line with the integration tests. It made more sense to move them to legacy instead of naming every new test file with the scenario suffix.

src/state/ingress_per_unit.py
swetha1654: you're right, it does wrap it in validationError. But i see that the pydantic validation is already done by the ingress_per_unit library. It raises a DataValidationError upon pydantic validation failure which is already taken care of here. So i will remove the validationError but will not catch the validationError as it won't be necessary.

PR #172, opened 2025-08-05
ISD-3914 Feat add https upstreams

dimaqq:


dimaqq:
This is work in progress, I'm happy to hand it over in this state.
Feel free to take some parts and not others.

templates/haproxy_route.cfg.j2
```diff
@@ -57,7 +57,7 @@ backend {{ backend.backend_name }}
     http-request {{ rewrite_configuration }}
 {% endfor %}
 {% for server in backend.servers %}
-    server {{ server.server_name }} {{ server.address }}:{{ server.port }} check inter {{ server.check.interval }}s rise {{ server.check.rise }} fall {{ server.check.fall }}
+    server {{ server.server_name }} {{ server.address }}:{{ server.port }} check inter {{ server.check.interval }}s rise {{ server.check.rise }} fall {{ server.check.fall }} {% if server.protocol == "https" %} ssl ca-file @system-ca {% endif %}
```
Thanhphan1147:
```suggestion
    server {{ server.server_name }} {{ server.address }}:{{ server.port }} check inter {{ server.check.interval }}s rise {{ server.check.rise }} fall {{ server.check.fall }} {% if server.protocol == "https" %} ssl ca-file @system-ca {% endif +%}
```

tests/unit/test_more_stuff.py
```diff
@@ -0,0 +1,92 @@
...
...AgIUN2mVv5+H7t0i3hAe06udfblztkswDQYJKoZIhvcNAQEL\\nBQAwLDEqMCgGA1UEAwwhc2VsZi1zaWduZWQtY2VydGlmaWNhdGVzLW9wZXJhdG9y\\nMB4XDTI1MDgwNTA1NDEwMFoXDTI2MDgwNTA1NDEwMFowLDEqMCgGA1UEAwwhc2Vs\\nZi1zaWduZWQtY2VydGlmaWNhdGVzLW9wZXJhdG9yMIIBIjANBgkqhkiG9w0BAQEF\\nAAOCAQ8AMIIBCgKCAQEAyJ9YZBAZrJfqhmMlbxX827B9Zu4cIAwLRLoLwcB+aucL\\nqZjpCn4qz4SQg/VXrEnKpNs18JtcvmehPwy8FaL6e2Ss3fb4BhbCQ5wwmU+Urrkb\\nPiK34kgn2h1bSRDe+lCuBS9Cw4Amrfgvm0qrWzVkhbhaUnoB+Q8K9oiOu41W+YJM\\n6BTw8CBOAledVoOf8Z1NyCfpFbAbSQaXPpaquUKQugQokzqBfLFU4LHOQNfhNlSb\\n5MZNctMXb6Wss+YAHMI0BgrHPQDswB20yOWVALZp8eQJ/hgV7+uYs044sfEfQsKQ\\nc2FiogqQl6VfJbAWX+AiN/elNTejiv/99/n+IFqAgQIDAQABo2cwZTAfBgNVHQ4E\\nGAQWBBT1vWAgwk/eadV65PGhrJ25/wJdIzAhBgNVHSMEGjAYgBYEFPW9YCDCT95p\\n1Xrk8aGsnbn/Al0jMA4GA1UdDwEB/wQEAwICpDAPBgNVHRMBAf8EBTADAQH/MA0G\\nCSqGSIb3DQEBCwUAA4IBAQAG5DoW4Y6878CaVBUTT6gO/YoS1gJGNew5eNyAbD4l\\nEI2PjICYm+XqISOtdneexeZxqBZ2sqllEmLZqWbFJgJZtjenBkQZC9zLdJKWr7GW\\nQqtZZUo4hZ2Ed0bBidZw4zCQZ+MiWad7hRQfHuHQyhZPPxbA7CzKEV12UN6dJfFd\\nw41a2EoXXl2ZEa4KiNUvMYrpPVW8y6uailjEb/MzId0LFiGiygOKoHsTYUqg77Vs\\nkcdFEAWJNpzqEMN/xecmEEEqmM4cWViphLnmahOJkmV16h89UR+n36STXF0SAt82\\nxMrQ6BDmFG+D+y+BBbs3wN35x++5CgPv/wzTqea7hSgu\\n-----END CERTIFICATE-----"]}]'
```
Thanhphan1147: This should probably be split into multiple lines

templates/haproxy_route.cfg.j2
dimaqq: Would you explain why, now that it's all on one line?

dimaqq:
I wasn't sure if @javierdelapuente is able to toggle draft/ready status for the PR, so I've put it into ready for review.

One unit test is still failing.

I've run into the same earlier, and my take is that some refactoring is called for, of the charm code, not test code, but that should rather be done in a separate PR, so maybe let's see if we can plug it here with a comment, or provide full set of data in the test.

src/charm.py
```diff
@@ -318,10 +318,13 @@ def _configure_legacy(self, charm_state: CharmState) -> None:
...
-            self.haproxy_route_tcp_provider,
-            typing.cast(typing.Optional[str], self.model.config.get("external-hostname")),
-            self._get_peer_units_address(),
+            haproxy_route=self.haproxy_route_provider,
+            haproxy_route_tcp=self.haproxy_route_tcp_provider,
+            external_hostname=typing.cast(
+                typing.Optional[str], self.model.config.get("external-hostname")
+            ),
+            peers=self._get_peer_units_address(),
+            ca_certs_configured=HAPROXY_CAS_FILE.exists(),
```
Thanhphan1147: should we also include `bool(self.recv_ca_cert.get_all_certificates())` to check if a CA cert was transferred?

src/charm.py
```diff
@@ -390,10 +393,11 @@ def _get_certificate_requests(self) -> typing.List[CertificateRequestAttributes]
...
                     HaproxyRouteRequirersInformation.from_provider(
-                        self.haproxy_route_provider,
-                        self.haproxy_route_tcp_provider,
-                        external_hostname,
-                        self._get_peer_units_address(),
+                        haproxy_route=self.haproxy_route_provider,
+                        haproxy_route_tcp=self.haproxy_route_tcp_provider,
+                        external_hostname=external_hostname,
+                        peers=self._get_peer_units_address(),
+                        ca_certs_configured=HAPROXY_CAS_FILE.exists(),
```
Thanhphan1147: should we also include bool(self.recv_ca_cert.get_all_certificates()) to check if a CA cert was transferred?

tests/integration/conftest.py
```diff
@@ -182,31 +183,34 @@ def any_charm_ingress_per_unit_requirer_fixture(
...
     )
+    with tempfile.NamedTemporaryFile(dir=".") as tf:
+        tf.write(src_overwrite.encode("utf-8"))
+        tf.flush()
+        juju.deploy(
+            "any-charm",
+            app=ANY_CHARM_HAPROXY_ROUTE_REQUIRER_APPLICATION,
+            channel="beta",
+            config={
+                "src-overwrite": f"@{tf.name}",
```
Thanhphan1147, line 210: nice!

tests/integration/haproxy_route_requirer.py
```diff
@@ -1,17 +1,35 @@
...
 """haproxy-route requirer source."""
 
-import pathlib
+import sys
 
-import apt  # type: ignore
-import ops
-from any_charm_base import AnyCharmBase  # type: ignore
-from haproxy_route import HaproxyRouteRequirer  # type: ignore
+# The last one is the dynamic modules. That way we get the new criptography
```
Thanhphan1147:
```suggestion
# The last one is the dynamic modules. That way we get the new cryptography
```

tests/integration/haproxy_route_requirer.py
```diff
@@ -32,40 +70,64 @@ def start_server(self):
...
+                RewriteEngine On
+                RewriteRule .* - [R=503,L]
+        </VirtualHost>
+        """
+        ssl_site_file = pathlib.Path("/etc/apache2/sites-available/anycharm-ssl.conf")
+        ssl_site_file.write_text(ssl_host, encoding="utf-8")
+        self._run_subprocess(["a2dissite", "000-default"])
+        self._run_subprocess(["a2ensite", "anycharm-ssl"])
+        self._run_subprocess(["a2enmod", "ssl", "rewrite"])
+        self._run_subprocess(["systemctl", "restart", "apache2"])
```
Thanhphan1147: would it be more clear to have something like this?

```python
cmd = (
  "a2dissite 000-default; "
  "a2ensite anycharm-ssl; "
  "a2enmod ssl rewrite; "
  "systemctl restart apache2"
)
self._run_subprocess(["/bin/bash", "-c", cmd])
```

Thanhphan1147 (approved):
LGTM, only a few comments. Thanks!

src/charm.py
javierdelapuente: yeah, I was not sure if to check the file of the get_all_certificates. As we commented that a file could be added manually I went this way.

Do you think it could be better to use `self.recv_ca_cert.get_all_certificates` instead? That should also imply the file exists (unless there is a big bug).

tests/integration/haproxy_route_requirer.py
javierdelapuente: Thanks for the suggestion! I changed it to a list of lists, so if some of the commands fail it raises.

src/charm.py
Thanhphan1147: I think we should discourage putting the cert manually so let's also check for the relation data for now. Wdyt?

Thanhphan1147 (approved):

.github/workflows/integration_test.yaml
```diff
@@ -13,3 +13,4 @@ jobs:
       self-hosted-runner: true
       charmcraft-channel: latest/edge
       modules: '["test_action.py","test_charm.py","test_config.py","test_cos.py","test_ha.py","test_haproxy_route.py","test_http_interface.py","test_ingress.py", "test_ingress_per_unit.py", "test_haproxy_route_tcp.py"]'
+      tmate-debug: true
```
swetha1654: is this needed?

swetha1654 (approved):

.github/workflows/integration_test.yaml
javierdelapuente: removed!

src/charm.py
javierdelapuente: done!

PR #173, opened 2025-08-05
ISD-3936 Add haproxy-route-tcp library implementation

Thanhphan1147:
Applicable spec: ISD-236 ( internal )

Add the haproxy-route-tcp library implementation with quite a few overlap with the `haproxy-route` library. Notable changes include:

1. removed references to HTTP attributes
2. More flexible TCP health checks and added specific types depending on workload
3. Added option to disable/enable tls termination and tls enforcement
4. replace unique attribute `service` to `port` 

The PR also added unit tests and update tox target.

### Library Changes

Added a new library

### Checklist

- [x] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [x] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [x] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [x] The documentation for charmhub is updated
- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)

PR #175, opened 2025-08-06
Feat: Add http-server-close

alithethird:
Applicable ticket: ISD-3955

### Overview

Adding the `http-server-close` argument to the haproxy config. When this argument is set it closes the connection after request.

### Checklist

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -86,12 +86,6 @@ config:
     vip:
       type: string
       description: Virtual IP address, used in active-passive ha mode.
-    http-server-close:
-      type: boolean
-      default: false
-      description: |
-          If set to true, the server will close the connection after each request.
-          This can be useful for certain applications that do not support keep-alive.
 
 actions:
   get-certificate:
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -86,6 +86,12 @@ config:
     vip:
       type: string
       description: Virtual IP address, used in active-passive ha mode.
+    http-server-close:
+      type: boolean
+      default: false
+      description: |
+          If set to true, the server will close the connection after each request.
+          This can be useful for certain applications that do not support keep-alive.
 
 actions:
   get-certificate:
--- a/lib/charms/haproxy/v1/haproxy_route.py
+++ b/lib/charms/haproxy/v1/haproxy_route.py
@@ -61,6 +61,7 @@ class SomeCharm(CharmBase):
         queue_timeout=<optional>,
         server_maxconn=<optional>,
         unit_address=<optional>,
+        http_server_close=<optional>,
     )
 
     # 2.To initialize the requirer with no parameters, i.e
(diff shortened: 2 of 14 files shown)


lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -148,7 +149,7 @@ def _on_haproxy_route_data_available(self, event: EventBase) -> None:
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 4
+LIBPATCH = 6
```
swetha1654, line 154: shouldn't this be 5?

src/haproxy.py
```diff
@@ -194,6 +194,8 @@ def _render_haproxy_config(self, template_file_path: str, context: dict) -> None
         )
         template = env.get_template(template_file_path)
         rendered = template.render(context)
+        logger.warning(f"@@@ Rendering HAProxy config: {rendered}")
+        logger.warning(f"@@@ HAProxy config context: {context}")
```
swetha1654: is this required?

lib/charms/haproxy/v1/haproxy_route.py
swetha1654, line 1: could you update the library docstring as well?

swetha1654 (commented):
Thanks for the contribution! Could you also add a unit test for the new configuration?

lib/charms/haproxy/v1/haproxy_route.py
alithethird: this [PR](https://github.com/canonical/haproxy-operator/pull/173#pullrequestreview-3090455151) is patch 5 and I imagine mine might merge later. If I get approvals before that PR I will update the libpatch before merging

src/haproxy.py
alithethird: Oh no :D Forgot them there. Thank you!

lib/charms/haproxy/v1/haproxy_route.py
alithethird: The docstring is updated already. Is there anything specific that I am missing?

seb4stien (commented):
Do we have a test to ensure that "option http-server-close" is not present if the config option is unset?

charmcraft.yaml
```diff
@@ -86,6 +86,12 @@ config:
     vip:
       type: string
       description: Virtual IP address, used in active-passive ha mode.
+    http-server-close:
+      type: boolean
+      default: false
+      description: |
+          If set to true, the server will close the connection after each request.
+          This can be useful for certain applications that do not support keep-alive.
```
Thanhphan1147: We should probably remove this for now and leave it for another PR

alithethird: Oh, I think this should not be here at all!

If I understand correctly, this setting can be different for different backends, so it should only be configured by `ingress-integrator`.

Is my understanding correct?

seb4stien (approved):

lib/charms/haproxy/v1/haproxy_route.py
swetha1654: oh my bad, i missed it.

swetha1654 (approved):

charmcraft.yaml
Thanhphan1147: yep!

PR #177, opened 2025-08-08
Isd 3937 render haproxy route tcp relation data

Thanhphan1147:
Add logic to aggregate haproxy-route-tcp relation data and transform them to a list of `HAProxyRouteTcpEndpoint` that gets rendered in a new template that extends the `haproxy_route.cfg.j2` template.

Fixes to cert handling logic

### Checklist


templates/haproxy_route_tcp.cfg.j2
```diff
@@ -0,0 +1,50 @@
...
+{% for option in endpoint.tcp_check_options %}
+    {{ option }}
+{% endfor %}
+
+{% for server in endpoint.servers %}
+    {{ render_server(server) | replace("\n", " ") }}
+{% endfor %}
+{% endfor %}
+
+{% endblock %}
```
alithethird: End of line

tests/integration/haproxy_route_requirer.py
```diff
@@ -0,0 +1,71 @@
...
+# See LICENSE file for licensing details.
+
+"""haproxy-route requirer source."""
+
+import pathlib
+
+import apt  # type: ignore
+import ops
+from any_charm_base import AnyCharmBase  # type: ignore
+from haproxy_route import HaproxyRouteRequirer  # type: ignore
```
alithethird, line 12: Why do we need these `# type: ignore`'s?

tests/integration/haproxy_route_requirer.py
```diff
@@ -0,0 +1,71 @@
...
+
+HAPROXY_ROUTE_RELATION = "require-haproxy-route"
+
+
+class AnyCharm(AnyCharmBase):
+    """haproxy-route requirer charm."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialize the requirer charm."""  # noqa
```
alithethird, line 21: Can you add a comment here for `# noqa`?

tests/integration/haproxy_route_tcp_requirer.py
```diff
@@ -0,0 +1,89 @@
...
+"""haproxy-route requirer source."""
+
+import logging
+import subprocess  # nosec
+import textwrap
+from pathlib import Path
+
+import ops
+from any_charm_base import AnyCharmBase  # type: ignore
+from haproxy_route_tcp import HaproxyRouteTcpRequirer, TCPHealthCheckType  # type: ignore
```
alithethird: Comment the ignore's please

alithethird (approved):
Just a couple nitpicks, mostly about commenting ignore lines.

src/charm.py
```diff
@@ -160,6 +163,12 @@ def __init__(self, *args: typing.Any):
         self.framework.observe(
             self.haproxy_route_provider.on.data_removed, self._on_config_changed
         )
+        self.framework.observe(
+            self.haproxy_route_tcp_provider.on.data_available, self._on_config_changed
+        )
+        self.framework.observe(
+            self.haproxy_route_tcp_provider.on.data_removed, self._on_config_changed
+        )
```
swetha1654, line 171: what happens when relation data changes?

lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -151,7 +151,7 @@ def _on_haproxy_route_data_available(self, event: EventBase) -> None:
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 6
+LIBPATCH = 7
```
arturo-seijas: Is this change intentional?

src/state/haproxy_route.py
```diff
@@ -238,13 +252,29 @@ def from_provider(
 
                 backends.append(backend)
 
+            tcp_endpoints: list[HAProxyRouteTcpEndpoint] = []
+            tcp_requirers: HaproxyRouteTcpRequirersData = haproxy_route_tcp.get_data(
+                haproxy_route_tcp.relations
+            )
+            relation_ids_with_invalid_data_tcp = tcp_requirers.relation_ids_with_invalid_data
+            for tcp_requirer in tcp_requirers.requirers_data:
+                if haproxy_route.relations and tcp_requirer.application_data.port in [80, 443]:
```
arturo-seijas, line 261: What if two requires request the same port? Don't we need to mark one as invalid?

src/state/haproxy_route_tcp.py
```diff
@@ -0,0 +1,134 @@
...
+    def from_haproxy_route_tcp_requirer_data(cls, provider: HaproxyRouteTcpRequirerData) -> "Self":
+        """Instantiate a HAProxyRouteTcpEndpoint class from the parent class.
+
+        Args:
+            provider: parent class.
+
+        Returns:
+            Self: The instantiated HAProxyRouteTcpEndpoint class.
+        """
+        return cls(**RootModel[HaproxyRouteTcpRequirerData](provider).model_dump())
```
arturo-seijas: I dont't like much the idea of deserializing with Pydantic, I feel we're introducing a point of dependency that we shouldn't need to

src/state/haproxy_route_tcp.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @property
+    def consistent_hashing(self) -> bool:
+        """Indicate if consistent hashing should be applied for this backend.
+
+        We make the assumption that the data have been validated by the library.
```
arturo-seijas: s/have/has

src/state/haproxy_route_tcp.py
```diff
@@ -0,0 +1,134 @@
...
+
+        Generates the appropriate HAProxy configuration lines for TCP health
+        checks based on the check type. For generic checks, creates send/expect
+        options. For database checks, includes the database user if specified.
+
+        Returns:
+            list[str]: List of HAProxy TCP check configuration options.
+        """
+        if check := self.application_data.check:
+            if check.check_type is None:
```
arturo-seijas: This can be a but simpler removing lines 121 and 121 and returning the value on 132 only `if check.check_type is not None`

src/charm.py
```diff
@@ -294,14 +304,27 @@ def _configure_legacy(self, charm_state: CharmState) -> None:
...
         haproxy_route_requirers_information = HaproxyRouteRequirersInformation.from_provider(
             self.haproxy_route_provider,
+            self.haproxy_route_tcp_provider,
             typing.cast(typing.Optional[str], self.model.config.get("external-hostname")),
             self._get_peer_units_address(),
         )
+        # We ONLY allow the charm to run with no certificate requested if:
+        # 1. there's only haproxy-route-tcp relations
+        # AND
+        # 2. All requirers must have tls passthrough or disable TLS termination
```
arturo-seijas: s/tls/TLS

src/charm.py
```diff
@@ -294,14 +304,27 @@ def _configure_legacy(self, charm_state: CharmState) -> None:
...
+        allow_no_certificates = (
```
arturo-seijas, line 317: Can you add unit tests covering this piece of code?

src/state/tls.py
```diff
@@ -85,24 +89,33 @@ def from_charm(
...
             charm: The haproxy charm.
             certificates: TLS certificates requirer class.
+            allow_no_certificates: If the charm can proceed without requesting any certificates.
 
         Raises:
             TLSNotReadyError: if the charm is not ready to handle TLS.
         """
         tls_requirer_integration = charm.model.get_relation(certificates.relationship_name)
-        if not certificates.certificate_requests:
+        if allow_no_certificates:
```
arturo-seijas, line 115: Can you add unit tests covering this piece of code?

src/state/charm_state.py
```diff
@@ -125,7 +128,9 @@ def _validate_state(
         is_ingress_related = bool(ingress_provider.relations)
         is_ingress_per_unit_related = bool(ingress_per_unit_provider.relations)
         is_legacy_related = bool(reverseproxy_requirer.relations)
-        is_haproxy_route_related = bool(haproxy_route_provider.relations)
+        is_haproxy_route_related = bool(
```
arturo-seijas, line 131: I don't see any unit tests changes for this change

src/state/haproxy_route.py
```diff
@@ -238,13 +252,29 @@ def from_provider(
 
                 backends.append(backend)
 
+            tcp_endpoints: list[HAProxyRouteTcpEndpoint] = []
+            tcp_requirers: HaproxyRouteTcpRequirersData = haproxy_route_tcp.get_data(
+                haproxy_route_tcp.relations
+            )
+            relation_ids_with_invalid_data_tcp = tcp_requirers.relation_ids_with_invalid_data
+            for tcp_requirer in tcp_requirers.requirers_data:
```
arturo-seijas, line 260: Can you unit test this change?

src/state/haproxy_route_tcp.py
```diff
@@ -0,0 +1,134 @@
+# Copyright 2025 Canonical Ltd.
```
arturo-seijas, line 1: I don't see any unit tests for this file

lib/charms/haproxy/v0/haproxy_route_tcp.py
```diff
@@ -0,0 +1,1738 @@
...
+        """Check if sni is configured but TLS is disabled.
+
+        Raises:
+            ValueError: If sni is configured and TLS is disabled.
+
+        Returns:
+            The validated model.
+        """
+        if not self.enforce_tls and self.sni is not None:
+            raise ValueError("setting sni and disabling TLS are mutually exclusive.")
```
arturo-seijas: If they are mutually exclusive the condition above is not ok, isn't it? That also doesn't match the docstring, what's the expected behavior here?

lib/charms/haproxy/v0/haproxy_route_tcp.py
```diff
@@ -0,0 +1,1738 @@
...
+        return self
+
+
+class HaproxyRouteTcpProviderAppData(_DatabagModel):
```
arturo-seijas: Why wrapping this single property in a class?

lib/charms/haproxy/v0/haproxy_route_tcp.py
```diff
@@ -0,0 +1,1738 @@
...
+    """haproxy-route provider databag schema.
+
+    Attributes:
+        endpoints: The list of proxied endpoints that maps to the backend.
+    """
+
+    endpoints: list[AnyUrl]
+
+
+class TcpRequirerUnitData(_DatabagModel):
```
arturo-seijas: Why wrapping this single property in a class?

lib/charms/haproxy/v0/haproxy_route_tcp.py
```diff
@@ -0,0 +1,1738 @@
...
+        """Check that requirers define unique ports.
+
+        Returns:
+            The validated model, with invalid relation IDs updated in
+                `self.relation_ids_with_invalid_data`
+        """
+        # Maybe the logic here can be optimized, we want to keep track of
+        # the relation IDs that request overlapping ports to ignore them during
+        # rendering of the haproxy configuration.
+        port_to_relation_ids: dict[int, list[int]] = {}
```
arturo-seijas: s/port_to_relation_ids/relation_ids_per_port

lib/charms/haproxy/v0/haproxy_route_tcp.py
```diff
@@ -0,0 +1,1738 @@
...
+        for requirer_data in self.requirers_data:
+            if not port_to_relation_ids.get(requirer_data.application_data.port):
```
arturo-seijas: You can use a defaultdict to simplify the logic

lib/charms/haproxy/v0/haproxy_route_tcp.py
```diff
@@ -0,0 +1,1738 @@
...
+
+class HaproxyRouteTcpDataRemovedEvent(EventBase):
+    """HaproxyRouteDataRemovedEvent custom event.
+
+    This event indicates that one of the endpoints was removed.
+    """
+
+
+class HaproxyRouteTcpProviderEvents(CharmEvents):
+    """List of events that the TLS Certificates requirer charm can leverage.
```
arturo-seijas: TLS Certificates?

lib/charms/haproxy/v0/haproxy_route_tcp.py
```diff
@@ -0,0 +1,1738 @@
...
+            queue_timeout=queue_timeout,
+            server_maxconn=server_maxconn,
+            ip_deny_list=ip_deny_list,
+            enforce_tls=enforce_tls,
+            tls_terminate=tls_terminate,
+        )
+        self._unit_address = unit_address
+
+        on = self.charm.on
+        self.framework.observe(on[self._relation_name].relation_created, self._configure)
```
arturo-seijas: Won't this lead to duplicated ready events when the relation is created? Iirc both a relation created and a relation changes will be triggered

lib/charms/haproxy/v0/haproxy_route_tcp.py
```diff
@@ -0,0 +1,1738 @@
...
+                If this is set to True the provider charm needs to also catch and handle the
+                thrown exception.
+        """
+        super().__init__(charm, relation_name)
+
+        self._relation_name = relation_name
+        self.charm = charm
+        self.raise_on_validation_error = raise_on_validation_error
+        on = self.charm.on
+        self.framework.observe(on[self._relation_name].relation_created, self._configure)
```
arturo-seijas: Won't this lead to duplicated ready events when the relation is created? Iirc both a relation created and a relation changes will be triggered

src/state/haproxy_route.py
Thanhphan1147: iirc that check should already be done in the library. For this specific check since we also need `haproxy-route` data we can only do it here at the charm level

lib/charms/haproxy/v0/haproxy_route_tcp.py
Thanhphan1147: This is the representation of the application databag. It's declared as a class so that we can use the `from_json` method of the child class `_DatabagModel` to populate the class using the relation data regardless of how many attributes there are in the relation data. It's a convention that we set to uniformly manage relation data across the lib, wdyt?

Thanhphan1147: Also this diff should be removed once #173 is merged.

lib/charms/haproxy/v0/haproxy_route_tcp.py
Thanhphan1147: As mentioned above

lib/charms/haproxy/v1/haproxy_route.py
Thanhphan1147: yep, it should be gone once #173 is merged

src/state/haproxy_route_tcp.py
Thanhphan1147: iirc `HaproxyRouteTcpRequirerData` is already a pydantic dataclass, so we are already depending on pydantic here.

lib/charms/haproxy/v1/haproxy_route.py
Thanhphan1147: my bad, didn't realize that this is the wrong lib. Will remove it.

tests/unit/legacy/test_state.py
swetha1654, line 1: we decided to use jubilant for any new tests added or existing test updates. Could you move this out of the legacy folder and convert it to jubilant?

Thanhphan1147: you mean `scenario`? I'd say that the unit tests here that deal with the charm state should not use either harness or scenario but instantiate the state class directly, will include this refactoring with the PR

tests/integration/haproxy_route_requirer.py
Thanhphan1147: since `haproxy_route` will exist only in the any-charm unit when we update its `src-overwrite` config. but here in vcs our linter will complain about those kind of imports

tests/integration/haproxy_route_requirer.py
Thanhphan1147: added!

src/state/haproxy_route_tcp.py
Thanhphan1147: I want to avoid nesting `if`s if possible so I did it like this, do you think I should remove it?

lib/charms/haproxy/v0/haproxy_route_tcp.py
Thanhphan1147: if we are not enforcing TLS then we can't guarantee to match ACLs based on the SNI, so if `self.sni` is set and `self.enforce_tls` if False then we should consider it an invalid configuration. Can you elaborate a bit more on what should I change here?

lib/charms/haproxy/v0/haproxy_route_tcp.py
Thanhphan1147: The ready event only indicate that there is data in the relation databag, but that data can also change between ready events, so i think we should fire the ready event every time and make no assumption on the content of the data available, wdyt?

lib/charms/haproxy/v0/haproxy_route_tcp.py
Thanhphan1147: similar to mentioned above.

src/charm.py
Thanhphan1147: Not sure I understood your question, can you elaborate a bit? The library wraps `relation_*` events and fire custom events (`data_available` and `data_removed`) and the charm implementing the library should listen on the custom events to know if it should start fetching the relation data.

templates/haproxy_route_tcp.cfg.j2
Thanhphan1147: updated!

tests/integration/haproxy_route_tcp_requirer.py
Thanhphan1147: updated!

lib/charms/haproxy/v0/haproxy_route_tcp.py
Thanhphan1147: updated!

lib/charms/haproxy/v0/haproxy_route_tcp.py
Thanhphan1147: updated, thanks!

lib/charms/haproxy/v0/haproxy_route_tcp.py
Thanhphan1147: updated!

src/state/haproxy_route_tcp.py
Thanhphan1147: updated!

src/charm.py
Thanhphan1147: updated!

src/state/tls.py
Thanhphan1147: added!

tests/unit/legacy/test_state.py
Thanhphan1147: updated

src/charm.py
Thanhphan1147: added

src/state/charm_state.py
Thanhphan1147: added

src/state/haproxy_route_tcp.py
Thanhphan1147: added

src/state/haproxy_route_tcp.py
arturo-seijas: The change wouldn't require nesting any ifs

lib/charms/haproxy/v0/haproxy_route_tcp.py
arturo-seijas: I'm not sure, but I find that the docstring, the code and the exception message are not consistent among each other

lib/charms/haproxy/v0/haproxy_route_tcp.py
arturo-seijas: What I was thinking is that the relation_created hook is redundant with the relation_changed, am I missing something?

src/state/haproxy_route_tcp.py
arturo-seijas: We are depending on pydantic for validation, but do we need to depend on it for serialization too? Why not using the json package instead, for instance, which is part of the stdlib?

Thanhphan1147: I don't think I follow here, `HaproxyRouteTcpRequirerData` is already a pydantic dataclass, so converting it to a model and calling `model_dump` should not introduce any extra dependencies. Can you give me a quick example of how would we do it with `json`?

src/state/haproxy_route.py
Thanhphan1147: added!

src/state/haproxy_route_tcp.py
Thanhphan1147: Ah I see what you mean, I updated the logic. Let me know if the changes are what you expect.

lib/charms/haproxy/v0/haproxy_route_tcp.py
Thanhphan1147: I'm not sure if there are cases where only one event would fire, but event if that's the case I still think we can afford hooking both here since the library does a preliminary validation at this level already so only valid relation data trigger a `ready` event, and if the data is already valid when we receive the `relation-created` event we should probably start processing that data already instead of waiting for a `relation-changed`. Though I'm not sure if that's even possible to have a complete relation data at the `relation-created` event (maybe the timing between the controller registering the relation data and the actual arrival of the `relation-created` event can cause such scenarios, but I'd say we should keep hooking both just to be on the safe side - and since data is validated before the custom event is fired I don't think we need to worry that much about performance or having too many events. wdyt?

arturo-seijas: A relation changed event will always be triggered after a relation created, so I feel like we increase the complexity by unnecessarily adding the relation created hook. I'm not thinking about performance but about overall complexity and ease of understanding the behavior

lib/charms/haproxy/v0/haproxy_route_tcp.py
Thanhphan1147: updated the exception message.

lib/charms/haproxy/v0/haproxy_route_tcp.py
Thanhphan1147: got it, I'll remove this for now.

Thanhphan1147: updated

src/state/haproxy_route_tcp.py
arturo-seijas: For instance, I see https://github.com/canonical/haproxy-operator/blob/main/lib/charms/haproxy/v0/haproxy_route_tcp.py#L233 using the json package to serialize so, why not doing the same or even using the parent class method?
Another thing I see here is that HAProxyRouteTcpEndpoint doesn't actually define any attributes, they are dynamically defined based on what gets dumped, and that defeats the purpose of parsing the data into a class.
Also, `cls(**RootModel[HaproxyRouteTcpRequirerData](provider).model_dump())` is quite difficult to reason about, so at least it will require a comment clarifying what this actauly does

Thanhphan1147: yes, the `HAProxyRouteTcpEndpoint` is basically a subclass of `HaproxyRouteTcpRequirerData` that do additional processing on the relation data to add haproxy-specific fields to make rendering the template easier, while the parent class (`HaproxyRouteTcpRequirerData`) is specific to the library and should not include haproxy-specific language. Can you let me know what would you prefer if we're not using a subclass?

About initialization/serialization I'm only looking to have a dict of attributes of the parent class to use them to initialize the subclass to avoid having to also change this method when adding new attributes, I'll update it to explicitly specify the attributes to avoid the confusion.

Thanhphan1147: updated

arturo-seijas (approved):

src/charm.py
Thanhphan1147: Closing this thread, we can resume the discussion internally and make changes in a follow-up PR if needed.

PR #180, opened 2025-08-13
Refine some release artifacts

seb4stien:
I reviewed the existing artifacts before adding new ones. These are just proposal, feedback is welcome.

Some notes taken during the process:
- It would help having a _template.yaml for the creation of new artifacts. It should be ignored by the generation.
- We should provide some guidelines on minor vs major.
- Do we recommend/align on something for the style of the description:
  - You can now deploy HAproxy with terraform vs
  - HAProxy can now be deployed with terraform

### DIFF
--- a/docs/release-notes/artifacts/pr0105.yaml
+++ b/docs/release-notes/artifacts/pr0105.yaml
@@ -10,5 +10,5 @@ changes:
       pr: https://github.com/canonical/haproxy-operator/pull/105
       related_doc: 
       related_issue: 
-    visibility: public
+    visibility: internal
     highlight: false
--- a/docs/release-notes/artifacts/pr0119.yaml
+++ b/docs/release-notes/artifacts/pr0119.yaml
@@ -5,7 +5,7 @@ changes:
   - title: Allow customization to configure port for HTTP check
     author: Thanhphan1147
     type: minor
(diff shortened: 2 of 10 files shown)


erinecon:
> Some notes taken during the process:
> 
> * It would help having a _template.yaml for the creation of new artifacts. It should be ignored by the generation.
> * We should provide some guidelines on minor vs major.
> * Do we recommend/align on something for the style of the description:
>   
>   * You can now deploy HAproxy with terraform vs
>   * HAProxy can now be deployed with terraform

Hi @seb4stien , thanks for looking over the artifacts and for your notes!

I definitely agree that including a  `_template.yaml` file would be useful for contributors. 

I think that having an engineer perspective on "minor" versus "major" features would be helpful for me here. From my perspective, a major feature or update would have significant consequences for the user, either in the way they interact with the charm or how the charm responds/react to usage. It would be hard for a user to ignore the existence of a major feature. On the other hand, a minor feature will be less noticed or impactful to the user experience. Does that make sense?

For major and minor updates, I think it's nice to phrase the changes from the user's perspective. The user can now use or operate the charm in a different way. I also think it's nice for us to avoid recommending passive voice in our documentation as much as possible. So I would lean toward recommending "You can now...".

erinecon (approved):
Changes look good to me -- thanks for checking the artifacts and updating them!

PR #181, opened 2025-08-13
Add missing PRs

seb4stien:
Something that is not clear to me is when to use "hidden" versus not creating an artifact for a given PR?

### DIFF
--- a/docs/release-notes/artifacts/pr0166.yaml
+++ b/docs/release-notes/artifacts/pr0166.yaml
@@ -6,7 +6,7 @@ changes:
     author: Thanhphan1147
     type: minor
     description: |
-      The `retry_interval` attribute is removed since there's no way to configure that in HAproxy.
+      The `retry_interval` attribute has been removed since there's no way to configure that in HAproxy.
       To configure the time between 2 retries we can use `timeout.connect`.
     urls: 
       pr: https://github.com/canonical/haproxy-operator/pull/166
--- a/docs/release-notes/artifacts/pr0166.yaml
+++ b/docs/release-notes/artifacts/pr0166.yaml
@@ -1,0 +1,16 @@
+# --- Release notes artifact ----
+
+schema_version: 1
+changes:
+  - title: Render retry configuration in template and remove retry_interval attribute
+    author: Thanhphan1147
+    type: minor
+    description: |
+      The `retry_interval` attribute is removed since there's no way to configure that in HAproxy.
+      To configure the time between 2 retries we can use `timeout.connect`.
+    urls: 
+      pr: https://github.com/canonical/haproxy-operator/pull/166
+      related_doc: 
+      related_issue: 
+    visibility: public
+    highlight: false
--- a/docs/release-notes/artifacts/pr0166.yaml
+++ b/docs/release-notes/artifacts/pr0166.yaml
@@ -2,7 +2,7 @@
 
 schema_version: 1
 changes:
-  - title: Render retry configuration in template and remove retry_interval attribute
+  - title: Rendered `retry` configuration in template and remove `retry_interval` attribute
     author: Thanhphan1147
     type: minor
     description: |
--- a/docs/release-notes/artifacts/pr0171.yaml
+++ b/docs/release-notes/artifacts/pr0171.yaml
@@ -6,7 +6,7 @@ changes:
     author: Thanhphan1147
     type: major
     description: |
-      Consistent hashing reduce redistribution of clients to another backend server when server is added/removed. On haproxy it will add 
+      Consistent hashing reduce redistribution of clients to another backend server when server is added/removed. On HAProxy it will add 
       ```
       hash-type consistent
       ```
--- a/docs/release-notes/artifacts/pr0171.yaml
+++ b/docs/release-notes/artifacts/pr0171.yaml
@@ -1,0 +1,19 @@
+# --- Release notes artifact ----
+
+schema_version: 1
+changes:
+  - title: Add consistent hashing support
+    author: Thanhphan1147
+    type: major
+    description: |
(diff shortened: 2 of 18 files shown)


srbouffard:
`visibility: hidden` may be in a grey-ish area vs simply not having a change artefact. I'm on the fence, but I am leaning towards keeping it.
The way I think of it, `visibility` field is really "release-visibility". The `change artefacts` documents the changes, no matter if these changes should or shouldn't be visible within a release.

Going a step further, `visibility: internal` could be added and it might fit better with some of the changes. We might have internal projects, which we still want to have the release flow applied to, and in which case showing internal changes is perfectly fine.

erinecon:
> `visibility: hidden` may be in a grey-ish area vs simply not having a change artefact. I'm on the fence, but I am leaning towards keeping it. The way I think of it, `visibility` field is really "release-visibility". The `change artefacts` documents the changes, no matter if these changes should or shouldn't be visible within a release.

I agree with @srbouffard here (about leaning toward keeping "hidden" changes). I think having the artifacts even for "hidden" changes are useful, especially for documentation changes. Benefits of having artifacts for "hidden" changes include:
* Ability to pick up on all contributors in a release, even for "hidden" changes. We don't have to look back to the repo history to make sure we've included all contributors in the release.
* We have the freedom to include previously "hidden" changes if we decide we want to, e.g., include documentation updates in a release note. 
* We have a more complete history of the changes in the repository, possibly enabling their usage for automating a changelog in the future.

docs/release-notes/artifacts/pr0166.yaml
```diff
@@ -0,0 +1,16 @@
+# --- Release notes artifact ----
+
+schema_version: 1
+changes:
+  - title: Render retry configuration in template and remove retry_interval attribute
```
erinecon: A question I had while making artifacts is: Do we want to use code-formatting in the headers of our release notes? Because I'm tempted to suggest:
```suggestion
  - title: Render `retry` configuration in template and remove `retry_interval` attribute
```

Another question I have is, do we want to recommend past tense for headers, like:
```suggestion
  - title: Rendered retry configuration in template and remove retry_interval attribute
```

docs/release-notes/artifacts/pr0166.yaml
```diff
@@ -0,0 +1,16 @@
...
+    author: Thanhphan1147
+    type: minor
+    description: |
+      The `retry_interval` attribute is removed since there's no way to configure that in HAproxy.
```
erinecon:
```suggestion
      The `retry_interval` attribute has been removed since there's no way to configure that in HAproxy.
```
Nitpick, but reads a bit better to me

docs/release-notes/artifacts/pr0176.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+schema_version: 1
+changes:
+  - title: Update tutorial to use ingress-configurator and local app
+    author: javierdelapuente
+    type: minor
+    description: |
+    urls: 
+      pr: https://github.com/canonical/haproxy-operator/pull/176
+      related_doc:
```
erinecon: Since Javi updated the tutorial for this entry, should we include a link to the tutorial here?

docs/release-notes/artifacts/pr0178.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+schema_version: 1
+changes:
+  - title: Add changelog
+    author: swetha1654
+    type: minor
+    description:
+    urls: 
+      pr: https://github.com/canonical/haproxy-operator/pull/178
+      related_doc:
```
erinecon:
```suggestion
      related_doc: https://github.com/canonical/haproxy-operator/blob/main/docs/changelog.md

```

docs/release-notes/artifacts/pr0179.yaml
```diff
@@ -0,0 +1,15 @@
+# --- Release notes artifact ----
+
+schema_version: 1
+changes:
+  - title: fix typo
```
erinecon: Since the `title` is very important for `bugfix` entries (as the script only uses the title in the rendered output), I think we should be more descriptive here. Something like:
```suggestion
  - title: Fixed a typo in the cookie fetch method in the load-balancing algorithm
```

docs/release-notes/artifacts/pr0171.yaml
```diff
@@ -0,0 +1,19 @@
+# --- Release notes artifact ----
+
+schema_version: 1
+changes:
+  - title: Add consistent hashing support
+    author: Thanhphan1147
+    type: major
+    description: |
+      Consistent hashing reduce redistribution of clients to another backend server when server is added/removed. On haproxy it will add
```
erinecon:
```suggestion
      Consistent hashing reduce redistribution of clients to another backend server when server is added/removed. On HAProxy it will add 
```

docs/release-notes/artifacts/pr0175.yaml
```diff
@@ -0,0 +1,15 @@
+# --- Release notes artifact ----
+
+schema_version: 1
+changes:
+  - title: Feat: Add http-server-close support
```
erinecon:
```suggestion
  - title: Added http-server-close support

```
I think we should probably remove or recommend against using "conventional commit" style language in the change artifact titles. WDYT?

erinecon (commented):
Thanks for adding these artifacts! One thought I had from looking at these artifacts is that we should probably recommend using past tense for change artifact titles, to establish consistency.

docs/release-notes/artifacts/pr0176.yaml
seb4stien: I added the one to github as we don't have it yet on charmhub.

erinecon (approved):
Thanks!

PR #190, opened 2025-08-29
Add release notes workflow

erinecon:
### Overview

Add release notes automation workflow.

### Rationale

With the work we've done to build up the data, now we can add in the workflow to prepare for the next release.

Templates for the change artifacts and release artifacts were updated or added.

The PR template checklist was updated to remind developers to add a change artifact to relevant PRs.

### Juju Events Changes

None

### Module Changes

None

### Library Changes

None

### Checklist

- [X] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [X] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [X] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [X] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)
- [ ] The [changelog](../docs/changelog.md) is updated with user-relevant changes in the format of [keep a changelog v1.1.0](https://keepachangelog.com/en/1.1.0/)

There's no documentation in this PR.
There are no user-relevant changes in this PR (only developer-relevant changes).


swetha1654 (approved):

swetha1654:
@erinecon can you explain why we need to manually create a change_artifact for every PR?

erinecon:
> can you explain why we need to manually create a change_artifact for every PR?

Hi @swetha1654 , my thought process is that creating the change artifact for each PR will split up the work and reduce excessive time and effort. When a contributor is working on their PR, they're already thinking about the change they're making, so that feels like an opportune time to create the change artifact.

We could create all the change artifacts when we're ready to publish a new set of release notes, but this may take more time and effort than necessary as someone would need to go back through the PR history and determine which of the PRs require change artifacts, and then gather the context to fill in the artifact keys.

What do you think?

swetha1654:
@erinecon I'm wondering if this change artifact can be generated automatically. It might become arduous if this file has to be manually populated for every PR, both trivial and non-trivial. I have two thoughts here:
1. Do we still need changelog if we're populating this? In my opinion, this is a much more elaborate version of a changelog. 
2. What do you think of populating this file automatically by reading the title and description of the PR? This of course would depend on the developer not leaving the description empty, but we could perhaps think about that later.

erinecon:
Hi @swetha1654 ,

The short answer is that we can look into automatically generating the change artifacts, but it's not a straightforward feature to implement and wouldn't happen until next cycle at the earliest. So in the meantime the change artifacts will need to be added manually (either in individual PRs or all at once when we're ready to publish release notes).

> 1. Do we still need changelog if we're populating this? In my opinion, this is a much more elaborate version of a changelog.

I think it's reasonable to replace the checklist item to update the changelog with the change artifacts instead. I actually think that automatically generating the changelog is a more straightforward task, and if I have time this cycle, I have a stretch item to look into this more.

> 2. What do you think of populating this file automatically by reading the title and description of the PR? This of course would depend on the developer not leaving the description empty, but we could perhaps think about that later.

This proposal is actually being discussed in another team, and the TA of that team wants to work on the tooling to generate the change artifacts in the way you described. I have doubts about whether this will work for our team for a few reasons:
* As you pointed out, the developer may leave description empty.
* We use a PR template that we would need to parse in order to grab the relevant material for the change artifact. Someone could modify or even remove the PR template entirely, and that would potentially break the parsing. 
* There are keys in the change artifacts that are not so easily determined by the title and description of the PR (e.g. relevant documentation/issues, whether the change is internal or should be hidden, whether we want to highlight the change, etc.). We could look into updating our PR templates to better assist with grabbing this information, but then we return to the scenario where a contributor modifies, ignores, or removes the template.

swetha1654:
> We use a PR template that we would need to parse in order to grab the relevant material for the change artifact. Someone could modify or even remove the PR template entirely, and that would potentially break the parsing.

One solution could be that we simply fail the workflow that generates the artifact from the PR, forcing the user to revert the PR template to the original one. But i think this should be a team level discussion.

Thanhphan1147 (approved):

PR #200, opened 2025-09-22
Catch up on change artifacts and add a new workflow

erinecon:
### Overview

Add change artifacts for PRs opened after the last release notes publication. Add a new GitHub workflow that checks whether non-bot PRs contain a change artifact. The action fails if a change artifact is not included in the pull request.

### Rationale

Need to catch up on change artifacts since there was no expectation to create a change artifact for PRs. The GitHub action will check for compliance and remind the contributor to record their work.

### Juju Events Changes

None

### Module Changes

None

### Library Changes

None

### Checklist

- [X] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [X] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [X] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [X] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)
- [X] The [changelog](../docs/changelog.md) is updated with user-relevant changes in the format of [keep a changelog v1.1.0](https://keepachangelog.com/en/1.1.0/)

There's no documentation in this PR to update.


.github/workflows/check_release_notes_artifact.yaml
```diff
@@ -0,0 +1,38 @@
+name: Check for release notes artifact
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
```
erinecon: Not really sure if this is necessary... seems like the default behavior (`opened, reopened, synchronize`) would be enough.

seb4stien: We usually don't specify the "types" indeed and just use: `pull_request:` with nothing after.

.github/workflows/check_release_notes_artifact.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+jobs:
```
seb4stien, line 6: I think that's a good starting point to avoid missing important changes, but we will probably have situations where a PR may not need a release artifact and/or when multiple PR refer to the same artefact.
I'm not sure how we should address these ones, especially the second case.
For the first one, maybe we could exclude "chore/" branches from the check?
cc @srbouffard if you have any idea.

seb4stien:
Thanks a lot for adding the missing artefacts.

.github/workflows/check_release_notes_artifact.yaml
seb4stien: That's a discussion we can probably move out of this PR and see what comes in the future.
Given the importance of haproxy, it's probably better to enforce a bit too much now, and relax constraints later.

seb4stien (approved):

.github/workflows/check_release_notes_artifact.yaml
erinecon: Thanks for confirming, I'll update to the default behavior!

.github/workflows/check_release_notes_artifact.yaml
erinecon: I agree that we should consider more edge cases and situations, so that this check does not add unnecessary noise to our PRs. 

I'll open a PR in platform-engineering-charm-template with this workflow and we can iterate further there

erinecon: https://github.com/canonical/platform-engineering-charm-template/pull/91

.github/workflows/check_release_notes_artifact.yaml
```diff
@@ -0,0 +1,37 @@
...
+          MATCHING=$(printf "%s\n" "${ADDED_FILES}" | grep -E '^docs/release-notes/artifacts/' || true)
+
+          if [ -z "${MATCHING}" ]; then
+            echo "ERROR: No new file added under docs/release-notes/artifacts/ in this PR."
+            echo "Please add a change artifact file that summarizes your pull request under docs/release-notes/artifacts/."
+            exit 1
+          fi
+
+          echo "Found release-notes artifact(s) added in PR:"
+          printf "%s\n" "${MATCHING}"
```
Thanhphan1147: Missing a newline here.

.github/workflows/check_release_notes_artifact.yaml
Thanhphan1147, line 1: The change artifact seems to be a yaml file, do we want to add basic structure validation to the action?

Thanhphan1147 (approved):
LGTM aside from a few comments

.github/workflows/check_release_notes_artifact.yaml
erinecon: Thanks for catching! I'll update

.github/workflows/check_release_notes_artifact.yaml
erinecon: That's a good idea, I'll update. Thanks!

PR #214, opened 2025-10-14
feat: add action get proxied endpoints

### LINKED ISSUE #80: Add an action to get proxied endpoints
### Enhancement Proposal

Could you please add an action in the charm that lists the proxied endpoints?
It would be something similar to [this](https://github.com/canonical/traefik-k8s-operator/blob/main/actions.yaml#L1) in Traefik.
This would be useful so requirers can check the ingress URL without the need to look into relation data or the need to implement a similar action in each requirer.

f-atwi:
Applicable spec: <link>

### Overview

Adds action `get-proxied-endpoints` to get the endpoints proxied by haproxy.
CLOSES #80

### Checklist


Thanhphan1147 (approved):
Lgtm, thanks!

src/charm.py
```diff
@@ -562,6 +562,49 @@ def _get_unit_address(self) -> typing.Optional[str]:
             return str(bind_address)
         return None
 
+    def _get_backend_proxied_endpoints(self, backend: HAProxyRouteBackend) -> list[str]:
+        """Get the list of proxied endpoints for a given backend.
+
+        Args:
+            backend: The HAProxyRouteBackend instance.
+        """
+        paths = backend.application_data.paths if backend.path_acl_required else [""]
```
arturo-seijas, line 571: Why `[""]` instead of `[]`?

src/charm.py
```diff
@@ -562,6 +562,49 @@ def _get_unit_address(self) -> typing.Optional[str]:
...
+            for backend in haproxy_route_requirers_information.backends:
+                if backend_name == backend.backend_name:
+                    event.set_results(
+                        {"endpoints": json.dumps(self._get_backend_proxied_endpoints(backend))}
+                    )
+                    return
+            event.fail(f'No backend with name "{backend_name}"')
+            return
+
+        proxied_endpoints: list[str] = [
```
arturo-seijas: This is a bit difficult to reason about IMO. I'd split it in two instructions. Also, this can probably be changed a bit to filter the backends by backend name and remove the block in L591

src/charm.py
```diff
@@ -562,6 +562,49 @@ def _get_unit_address(self) -> typing.Optional[str]:
...
+            proxied_endpoint
+            for backend in haproxy_route_requirers_information.backends
+            for proxied_endpoint in self._get_backend_proxied_endpoints(backend)
+        ]
+        event.set_results({"endpoints": json.dumps(proxied_endpoints)})
```
arturo-seijas, line 602: What if proxies endpoints is empty? If empty is a valid result, it should be a valid result to when filtering by backend name

src/charm.py
```diff
@@ -562,6 +562,49 @@ def _get_unit_address(self) -> typing.Optional[str]:
...
+
+        Args:
+            backend: The HAProxyRouteBackend instance.
+        """
+        paths = backend.application_data.paths if backend.path_acl_required else [""]
+        return [
+            f"https://{hostname}/{path}" for hostname in backend.hostname_acls for path in paths
+        ]
+
+    def _on_get_proxied_endpoints_action(self, event: ActionEvent) -> None:
```
arturo-seijas, line 576: I'm missing a unit test for the action

src/charm.py
f-atwi: the endpoints are a combination of all the hostnames and paths
if there is no paths, then the root path is used
otherwise there would be no endpoints if there were no paths

src/charm.py
f-atwi: I have simplified the code but the logic was kept the same. Let me know if this is the desired behaviour.
- if no backend filter is provided, then all the proxied endpoints are return (even if there are no haproxy-route relations resulting in an empty list)
- if backend filter is provided, then the proxied endpoints for that backend are returned (if there are no proxied endpoints associated with that backend, it would still return an empty list, though I do not think this would happen since we would face other issues if the related charm does not have a hostname)
- if we specify a backend filter with a backend name that does not exists, we are failing the event. The reasoning for this is to distiguish between not having proxied endpoints for that backend and the backend not existing in the first place. For example if a typo was given in the backend_name, you wouldn't want to give an answer.

I think the last point is the contested point. Let me know if this is desired or I should switch to returning an empty list in this case

src/charm.py
f-atwi: i have simplified the code, it is a bit harder to merge all together if I continue to have the event.fail case.
My reply to the next comment describes why I opted for a fail. Let me know whether to keep it or no.

tests/unit/test_actions.py
```diff
@@ -0,0 +1,163 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Unit tests for the haproxy charm actions."""
```
arturo-seijas: Please, add this in a test_charm.py file so that every source files matches exactly one test file

src/charm.py
arturo-seijas: As it is now, an empty results in  an error if `backend_name` is set and in an empty result if it's not. I wouldn't return an error for the first case, an empty result should be valid (L597)

src/charm.py
```diff
@@ -587,22 +587,23 @@ def _on_get_proxied_endpoints_action(self, event: ActionEvent) -> None:
...
+                if backend.backend_name == backend_name
+            ]
+            if not backends:
+                event.fail(f'No backend with name "{backend_name}"')
+                return
+        else:
+            backends = haproxy_route_requirers_information.backends
+
+        proxied_endpoints = []
+        for backend in backends:
```
arturo-seijas: This can be a comprehension list

arturo-seijas (approved):

erinecon (approved):
Thanks for adding documentation for the new action in the YAML file :D I think it looks clear, and I hope that the autogenerated documentation in Charmhub will be enough for users to understand what this action does. If necessary, we can add a dedicated how-to guide later.

PR #225, opened 2025-11-12
chore: migrate to uv+ruff

weiiwang01:
### Overview

Migrate the Python project to use uv, tox-uv, and ruff for building, testing and linting.

The changes include:

* Merge `requirements.txt` into `pyproject.toml`
* Migrate `tox.ini` to `tox.toml`
* Replace `tox` with `tox-uv`
* Use `ruff` for linting and formatting, replacing `flake8`, `pylint`, `black`, and `isort`
* Fix some linting issue discovered by `ruff`

### Checklist

- [x] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [x] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [x] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [x] The documentation for charmhub is updated
- [x] A [change artifact](../docs/release-notes/template/docs/release-notes/template/_change-artifact-template.yaml) was added for user-relevant changes in `docs/release-notes/artifacts`. 
- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)
- [x] The [changelog](../docs/changelog.md) is updated with user-relevant changes in the format of [keep a changelog v1.1.0](https://keepachangelog.com/en/1.1.0/)

PR #230, opened 2025-11-12
Allow enabling http for haproxy route backend

Thanhphan1147:
Blocked until #225 is merged. Add a new `allow_http` attribute to allow disabling mandatory HTTPS redirection for backends. Add logic to build the required ACL and rendering logic in the j2 template.

### Checklist


src/state/haproxy_route.py
```diff
@@ -310,16 +310,34 @@ def check_backend_paths(self) -> Self:
...
         if len(requirers_hostnames) != len(set(requirers_hostnames)):
             logger.warning(
                 "Requirers defined hostname(s) that map to multiple backends."
-                "This can cause unintended behaviours."
+                "This can cause unintended behaviors."
             )
         return self
 
+    @property
+    def allow_http_acls(self) -> list[str]:
```
arturo-seijas: I would rename this method since it includes both allow and deny ACLs

arturo-seijas: This approach feels a bit inconsistent with how the rest of the ACLs are handled, which is more template based. See templates/haproxy_route.cfg.j2:13. I think it makes sense to either handle do this in the template or extract the logic to the Python code for the backend ACLs

tests/integration/test_haproxy_route.py
```diff
@@ -132,6 +132,7 @@ def test_haproxy_route_protocol_https(
                         "load_balancing_consistent_hashing": True,
                         "http_server_close": True,
                         "protocol": "https",
+                        "allow_http": True,
```
arturo-seijas, line 135: I think it makes sense to test the scenario where `allow_http` is false too. Not sure if that's already there

templates/haproxy_route.cfg.j2
```diff
@@ -6,7 +6,7 @@ frontend haproxy
     bind [::]:80 v4v6
     bind [::]:443 v4v6 ssl crt {{ haproxy_crt_dir }}
     # Redirect HTTP to HTTPS
-    http-request redirect scheme https unless { ssl_fc }
+    http-request redirect scheme https unless { ssl_fc } {% for allow_http_acl in allow_http_acls %} || {{ allow_http_acl }}{% endfor %}
```
gregory-schiano: To investigate:
Might be clearer to create a redirect line per pair of host/path, but we need to reverse the logic and use `if` instead of `unless`
The default behavior should fallback to redirect still

tests/unit/test_haproxy_route.py
```diff
@@ -0,0 +1,113 @@
...
+                    units_data=[
+                        typing.cast(
+                            RequirerUnitData, RequirerUnitData.from_dict({"address": "10.0.0.1"})
+                        )
+                    ],
+                ),
+            ],
+            relation_ids_with_invalid_data=[],
+        )
+    )
```
gregory-schiano: Should reduce the level of nesting by using local variables

Thanhphan1147: Updated!

tests/integration/test_haproxy_route.py
Thanhphan1147: The default value for that attribute is false so I think it's already tested by the other integration tests

src/state/haproxy_route.py
Thanhphan1147: I pre-rendered the ACLs here directly because they are very complex if we were to render it in the template ( we need to group the hostname / paths / deny-paths per backend that configured `allow_http`. I want to also eventually shift some of the more complex rendering logic from the template to the charm code, so this can serve as a good starting point.

templates/haproxy_route.cfg.j2
Thanhphan1147: I looked at this for a bit and it'm still not sure how I can split this into multiple lines. So I'll keep it like this for now and make a follow-up PR if needed.

src/state/haproxy_route.py
Thanhphan1147: I updated the name to avoid the confusion

arturo-seijas (approved):

alithethird (approved):

PR #243, opened 2025-11-19
update terraform module to support keepalived, update docs

Thanhphan1147:
Update TF module to provide support for keepalived, add validation rules to ensure mutual exclusivity between the 2 options.

### Checklist

### DIFF
--- a/terraform/product/variables.tf
+++ b/terraform/product/variables.tf
@@ -1,7 +1,7 @@
 # Copyright 2025 Canonical Ltd.
 # See LICENSE file for licensing details.
 variable "model_uuid" {
-  description = "ID of the model to deploy to, takes priority over model + model_owner"
+  description = "ID of the model to deploy to"
   type        = string
   default     = ""
 }
--- a/terraform/product/variables.tf
+++ b/terraform/product/variables.tf
@@ -6,18 +6,6 @@ variable "model_uuid" {
   default     = ""
 }
 
-variable "model" {
-  description = "Reference to the Juju model to deploy application to."
-  type        = string
-  default     = ""
-}
-
-variable "model_owner" {
-  description = "ID of the model owner, used in conjunction with model name."
-  type        = string
-  default     = "admin"
-}
-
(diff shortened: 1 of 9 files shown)


gregory-schiano (commented):
Update `versions.tf` to pin to `>=1.0`

Thanhphan1147:
updated the version to pin to `~> 1.0`

DeeKay3 (approved):

terraform/product/variables.tf
```diff
@@ -1,9 +1,9 @@
 # Copyright 2025 Canonical Ltd.
 # See LICENSE file for licensing details.
-
-variable "model" {
-  description = "Reference to the Juju model to deploy application to."
+variable "model_uuid" {
+  description = "ID of the model to deploy to, takes priority over model + model_owner"
```
javierdelapuente: There is no model/model_owner, so the comment does not look correct.

terraform/product/variables.tf
```diff
@@ -25,12 +25,25 @@ variable "hacluster" {
     config   = optional(map(string), {})
     revision = optional(number, null)
   })
-  default = {}
+  default = null
```
javierdelapuente, line 28: Just thinking out loud. Would it be clearer an extra var use_keepalived/use_haproxy instead?

weiiwang01 (approved):

seb4stien (approved):

terraform/product/variables.tf
Thanhphan1147: Thanks, I updated it!

terraform/product/variables.tf
Thanhphan1147: I think in the user-facing one we want a simple interface while in the internal module we have that extra arg to simplify our logic. It's would be strange if they set `use_keepalived=True` but set `hacluster={}` I think

Thanhphan1147: closing the comment for now, I'll continue this discussion internally If needed

PR #255, opened 2025-11-23
Skip initialization of the TLSInformation state component if `allow_no_certificates` is set to True

Thanhphan1147:
return None in `TLSInformation.from_charm` if `allow_no_certificates` is true, update charm code and tests.

### Checklist

### DIFF
--- a/docs/release-notes/artifacts/pr0255.yaml
+++ b/docs/release-notes/artifacts/pr0255.yaml
@@ -1,0 +1,14 @@
+schema_version: 1
+changes:
+  - title: Handle optional TLS information in charm and state
+    author: Phan Trung Thanh
+    type: bugfix
+    description: |
+      Allow TLSInformation.from_charm to return None when certificates are not required,
+      and handle this case in the charm code to avoid errors.
+    urls:
+      pr: https://github.com/canonical/haproxy-operator/pull/255
+      related_doc:
+      related_issue:
+    visibility: hidden
+    highlight: false
--- a/docs/release-notes/artifacts/pr0255.yaml
+++ b/docs/release-notes/artifacts/pr0255.yaml
@@ -1,7 +1,7 @@
 schema_version: 1
 changes:
   - title: Handle optional TLS information in charm and state
-    author: Phan Trung Thanh
+    author: tphan025
     type: bugfix
     description: |
       Allow TLSInformation.from_charm to return None when certificates are not required,
--- a/src/charm.py
+++ b/src/charm.py
@@ -265,9 +265,10 @@ class HAProxyCharm(ops.CharmBase):
             case ProxyMode.HAPROXY_ROUTE:
                 self._configure_haproxy_route(charm_state)
             case _:
-                if self.model.get_relation(TLS_CERT_RELATION):
-                    # Reconcile certificates in case the certificates relation is present
-                    tls_information = TLSInformation.from_charm(self, self.certificates)
+                # Reconcile certificates in case the certificates relation is present
+                if (
+                    tls_information := TLSInformation.from_charm(self, self.certificates)
+                ) and self.model.get_relation(TLS_CERT_RELATION):
                     self._tls.certificate_available(tls_information)
 
                 self.unit.set_ports(80)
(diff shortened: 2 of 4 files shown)


src/charm.py
```diff
@@ -265,9 +265,10 @@ def _reconcile(self) -> None:
...
                 self._configure_haproxy_route(charm_state)
             case _:
-                if self.model.get_relation(TLS_CERT_RELATION):
-                    # Reconcile certificates in case the certificates relation is present
-                    tls_information = TLSInformation.from_charm(self, self.certificates)
+                # Reconcile certificates in case the certificates relation is present
+                if (
+                    tls_information := TLSInformation.from_charm(self, self.certificates)
+                ) and self.model.get_relation(TLS_CERT_RELATION):
                     self._tls.certificate_available(tls_information)
```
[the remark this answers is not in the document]
Thanhphan1147: updated

DeeKay3 (approved):

src/charm.py
```diff
@@ -280,25 +281,26 @@ def _configure_ingress(
         requirer_class: IngressRequirersInformation | IngressPerUnitRequirersInformation,
     ) -> None:
         """Configure the ingress or ingress-per-unit relation."""
-        tls_information = TLSInformation.from_charm(self, self.certificates)
-        self._tls.certificate_available(tls_information)
+        if tls_information := TLSInformation.from_charm(self, self.certificates):
```
arturo-seijas, line 284: Don't you need `allow_no_certificates` here or the cert is mandatory in this case?

src/charm.py
```diff
@@ -265,9 +265,10 @@ def _reconcile(self) -> None:
             case ProxyMode.HAPROXY_ROUTE:
                 self._configure_haproxy_route(charm_state)
             case _:
-                if self.model.get_relation(TLS_CERT_RELATION):
-                    # Reconcile certificates in case the certificates relation is present
-                    tls_information = TLSInformation.from_charm(self, self.certificates)
+                # Reconcile certificates in case the certificates relation is present
+                if self.model.get_relation(TLS_CERT_RELATION) and (
```
arturo-seijas, line 269: Don't you need `allow_no_certificates` here?

src/charm.py
```diff
@@ -453,8 +457,9 @@ def _on_ca_certificates_removed(self, _: CertificatesRemovedEvent) -> None:
     def _on_ingress_per_unit_data_provided(self, _: IngressDataReadyEvent) -> None:
         """Handle the data-provided event for ingress-per-unit."""
         self._reconcile()
-        if self.unit.is_leader():
-            tls_information = TLSInformation.from_charm(self, self.certificates)
+        if self.unit.is_leader() and (
```
arturo-seijas, line 460: Is the certificate mandatory for the ingress per unit too?

src/charm.py
Thanhphan1147: I set False as the default value so that it doesn't have to be specified when not needed

src/charm.py
Thanhphan1147: I set False as the default value so that it doesn't have to be specified when not needed

arturo-seijas (commented):
The changes on `charm.py` are not unit tested

src/charm.py
Thanhphan1147: Yes, the cert is required for everything except when you are doing plain TCP and only plain TCP

src/charm.py
arturo-seijas: That means `LSInformation.from_charm`  is not nullable here, so why checking if it isn't none? Also, is unclear to me when `allow_no_certificates` is supposed to take effect and when it isn't. Is the behavior documented somewhere?

src/charm.py
arturo-seijas: Same as above, why checking if it's not none if it is required?

src/state/tls.py
```diff
@@ -53,7 +53,7 @@ def from_charm(
         charm: ops.CharmBase,
         certificates: TLSCertificatesRequiresV4,
         allow_no_certificates: bool = False,
-    ) -> "TLSInformation":
+    ) -> "TLSInformation | None":
```
seb4stien, line 56: @Thanhphan1147 I may be lacking context, but I find the approach a bit confusing: wouldn't it be clearer from the developper perspective to have an explicit "if" in the code before calling the TLSInformation.from_charm instead of hiding the magic in it?

src/state/tls.py
```diff
@@ -112,10 +115,8 @@ def validate(
             TLSNotReadyError: if the charm is not ready to handle TLS.
         """
         tls_requirer_integration = charm.model.get_relation(certificates.relationship_name)
-        if allow_no_certificates:
-            logger.warning("Allowing load balancing without requesting TLS certificates.")
 
-        if not certificates.certificate_requests and not allow_no_certificates:
```
seb4stien, line 118: I found this approach clearer. However "not allow_no_certificaties" should come first as things are tested in order.

docs/release-notes/artifacts/pr0255.yaml
```diff
@@ -0,0 +1,14 @@
+schema_version: 1
+changes:
+  - title: Handle optional TLS information in charm and state
+    author: tphan025
+    type: bugfix
+    description: |
+      Allow TLSInformation.from_charm to return None when certificates are not required,
+      and handle this case in the charm code to avoid errors.
```
seb4stien, line 8: We should focus on user experience rather than implementation details in the release notes. Maybe something like:

```suggestion
      Fixing a bug on how certificates where handled when not required (for TCP routes).
```

Thanhphan1147:
closing as duplicate with #304

PR #275, opened 2025-12-04
run reset-failed everytime we manually restart the haproxy-spoe-auth service

Thanhphan1147:
This is because this snap doesn't support reloading and the systemd service created by snapd uses the default restart limit of 5 every 10 seconds, so it's quite easy to hit the restart limit if events come too quickly, so we do this for "intentional" restarts to not hit the restart limit ( note that this is not done for on-failure restarts so we still have a limit in that case )

### Checklist

- [x] A [change artifact](../docs/release-notes/template/docs/release-notes/template/_change-artifact-template.yaml) was added for user-relevant changes in `docs/release-notes/artifacts`. If this PR does not require a change artifact, the PR has been tagged with `no-release-note`. 

### DIFF
--- a/haproxy-spoe-auth-operator/src/haproxy_spoe_auth_service.py
+++ b/haproxy-spoe-auth-operator/src/haproxy_spoe_auth_service.py
@@ -4,6 +4,10 @@
 """The haproxy-spoe-auth service module."""
 
 import logging
+
+# We silence this rule because subprocess call is only for
+# resetting the service restart counter and no user input is parsed
+import subprocess  # nosec B404
 from pathlib import Path
 
 from charmlibs import snap
@@ -20,6 +24,7 @@ OIDC_HEALTHCHECK_PATH = "/health"
 OIDC_LOGOUT_PATH = "/oauth2/logout"
 SNAP_CHANNEL = "latest/edge"
 SNAP_NAME = "haproxy-spoe-auth"
+SNAP_SERVICE_NAME = "snap.haproxy-spoe-auth.haproxy-spoe-auth.service"
 SPOP_PORT = 8081
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


haproxy-spoe-auth-operator/src/haproxy_spoe_auth_service.py
```diff
@@ -4,6 +4,10 @@
 """The haproxy-spoe-auth service module."""
 
 import logging
+
+# We silence this rule because subprocess call is only for
+# resetting the service restart counter and no user input is parsed
+import subprocess  # nosec B404
```
swetha1654: can you use the error name instead of the error code?

haproxy-spoe-auth-operator/src/haproxy_spoe_auth_service.py
```diff
@@ -72,8 +76,12 @@ def reconcile(self, charm_state: CharmState, oauth_information: OauthInformation
         try:
             self._render_config(charm_state, oauth_information)
             self.haproxy_spoe_auth_snap.restart(reload=True)
-        except snap.SnapError as exc:
-            raise SpoeAuthServiceConfigError(f"Failed to reconcile service: {exc}") from exc
+            # Ignore bandit rule as this is only used to reset the service restart counter.
+            subprocess.check_call(["/usr/bin/systemctl", "reset-failed", SNAP_SERVICE_NAME])  # nosec B603
```
swetha1654: may i know why we're resetting the restart counter?

Thanhphan1147: We do this because this snap doesn't support reloading and the systemd service created by snapd uses the default restart limit of 5 every 10 seconds, so it's quite easy to hit the restart limit if events come too quickly, so we do this for "intentional" restarts to not hit the restart limit ( note that this is not done for `on-failure` restarts so we still have a limit in that case )

haproxy-spoe-auth-operator/src/haproxy_spoe_auth_service.py
Thanhphan1147: thanks, updated

arturo-seijas (approved):

swetha1654 (approved):

seb4stien (approved):

PR #279, opened 2025-12-09
Support for spoe-auth in Haproxy. 

javierdelapuente:
Applicable spec: <link>

### Overview

This PR implements the spoe-auth relation in the haproxy charm.

The following are the main changes and decisions taken:
- It implements the spoe-auth. A few changes have been made over the original interface:
  - The variables var_authenticated and var_redirect_url have been split into scope and var name. The prefix is something relative to HAproxy and does not belong to the interface.
  - Events available and removed have been removed from the interface. As discussed, HAproxy has to be aware of an invalid relation to block the charm (to avoid possible security problems after all is configured)/
- The charm will block if the spoe relation is invalid. If something changes after the charm is properly configured, that may mean that the host protected will not be unprotected by mistake.
- A new integration test has been added that is run at the root of the repository and has access to both the haproxy and the haproxy-spoe-auth charm. This integration test deploys three hosts using haproxy-route (anycharm), two of them protected and one unprotected. Canonical IDP is used for this test.

Pending to this PR, a following one should be created for possible documentation (if this is required) and to renovate all integration tests.

### Checklist

- [x] A [change artifact](../docs/release-notes/template/docs/release-notes/template/_change-artifact-template.yaml) was added for user-relevant changes in `docs/release-notes/artifacts`. If this PR does not require a change artifact, the PR has been tagged with `no-release-note`. 
- [ ] The [changelog](../docs/changelog.md) is updated with user-relevant changes in the format of [keep a changelog v1.1.0](https://keepachangelog.com/en/1.1.0/)


haproxy-operator/src/state/spoe_auth.py
```diff
@@ -0,0 +1,71 @@
...
+            except (DataValidationError, SpoeAuthInvalidRelationDataError) as ex:
+                raise SpoeAuthValidationError from ex
+
+            try:
+                requirer_units_data = spoe_auth_requirer.get_provider_unit_data(relation)
+            except DataValidationError as ex:
+                raise SpoeAuthValidationError from ex
+
+            unit_addresses = [unit_data.address for unit_data in requirer_units_data]
+            spoe_auth_information = cls(
```
javierdelapuente, line 69: I am letting the exception of any validation in the relation propagate as a config error, so the charm gets into blocked state. I expect that we want to block when there is an error with this, as it could have security implications.

haproxy-operator/lib/charms/haproxy/v0/spoe_auth.py
```diff
@@ -447,10 +463,18 @@ def relation(self) -> Relation | None:
...
+        """The list of relations associated with this relation_name.
+
+        Returns:
+            The list of relations.
+        """
+        return list(self.charm.model.relations[self.relation_name])
+
     def _configure(self, _: EventBase) -> None:
         """Handle relation changed events."""
-        if self.is_available():
```
javierdelapuente, line 452: If we work this way, in the main charm we will have no information when the relation is in bad shape (a change that make the relation break, a starting relation...).

docs/release-notes/artifacts/pr0279.yaml
```diff
@@ -0,0 +1,13 @@
+schema_version: 1
+changes:
+  - title: Implement spoe-auth relation in HAproxy
```
seb4stien:
```suggestion
  - title: Implemented spoe-auth relation in HAproxy
```

haproxy-operator/templates/haproxy_route.cfg.j2
```diff
@@ -76,5 +87,32 @@ backend {{ backend.backend_name }}
...
+    mode http
+    balance roundrobin
+    http-request redirect location %[var({{ spoe_auth_info.var_redirect_url_scope }}.auth_{{ spoe_auth_info.id }}.{{ spoe_auth_info.var_redirect_url }})]
+
+backend spoe_oidc_callback_{{ spoe_auth_info.id }}
+    mode http
+    balance roundrobin
+    option tcp-check
+    timeout connect 5s
+    timeout server  3m
```
seb4stien: Isn't it a big long? Under load couldn't it accumulates a lot of open connexions and create some kind of DOS?
However, I don't expect SPOE to be used on heavy loads, so that's probably not an issue.

haproxy-operator/templates/haproxy_route.cfg.j2
```diff
@@ -76,5 +87,32 @@ backend {{ backend.backend_name }}
...
+{% for agent_address in spoe_auth_info.unit_addresses %}
+    server {{ agent_address }} {{ agent_address }}:{{ spoe_auth_info.oidc_callback_port }} check
+{% endfor %}
+
+backend spop_auth_{{ spoe_auth_info.id }}
+    mode tcp
+    balance roundrobin
+    option spop-check
+    timeout connect 5s
+    timeout server  3m
```
seb4stien: Same as above.

haproxy-operator/templates/haproxy_route.cfg.j2
javierdelapuente: I  put the same as in the example in the spoe repository.

Changed to 1m (I do not expect checking the token should take longer than that even in the worst case)....

haproxy-operator/templates/haproxy_route.cfg.j2
javierdelapuente: Updated to 1m

tests/integration/conftest.py
```diff
@@ -0,0 +1,386 @@
...
+):
+    """The haproxy charm configured and integrated with TLS provider."""
+    if pytestconfig.getoption("--no-deploy") and "haproxy" in lxd_juju.status().apps:
+        return application
+    lxd_juju.config(application, {"external-hostname": TEST_EXTERNAL_HOSTNAME_CONFIG})
+    lxd_juju.integrate(
+        f"{application}:certificates",
+        f"{certificate_provider_application}:certificates",
+    )
+    return application
```
seb4stien, line 144: Shouldn't we "yield" and do the cleanup afterwards to ensure the CI is not blocked due to some remaining resources?

tests/integration/conftest.py
```diff
@@ -0,0 +1,386 @@
...
+    k8s_juju.integrate("traefik-admin:ingress", "hydra:admin-ingress")
+    k8s_juju.integrate("traefik-admin:ingress", "kratos:admin-ingress")
+    k8s_juju.integrate("traefik-public:ingress", "hydra:public-ingress")
+    k8s_juju.integrate("traefik-public:ingress", "kratos:public-ingress")
+    k8s_juju.integrate(
+        "traefik-public:ingress", "identity-platform-login-ui-operator:ingress"
+    )
+
+    k8s_juju.config("kratos", {"enforce_mfa": False})
+    k8s_juju.offer(app=f"{k8s_juju.model}.hydra", endpoint="oauth")
```
seb4stien, line 259: Similar to previous comment: shouldn't we do some cleanup at the end of the tests?

seb4stien (approved):

tests/integration/conftest.py
javierdelapuente: The CI destroys the controller (with juju destroy-controller) in the post setup. If there is a bug somewhere (a charm stuck in error), the ci will get stuck, but because there is an error.

This is "wrongly" a module fixture, and probably fits better the "session" scope, but as we run the CI in modules, we have it that way everywhere. As session fixtures are reused for all the tests, I do not think it is necessary to clean them up (and can be reused between test invocations).

Said that, I haven't cleaned up any fixture, and there are fixtures that can be scoped to function. The reason is that I only have one test, and this way it is easier to check a test if it fails, but once another test is created, the clean up should be implemented.

tests/integration/conftest.py
javierdelapuente: See previous answer. I would also scope this fixture as session.

However, if you think it is interesting to change the fixtures that should be scoped as "function" (haproxy_spoe_auth_deployer and any_charm_haproxy_route_deployer), I could review it in this or a following PR. Otherwise I would leave it unimplemented (and should be done when another test is created).

tests/integration/conftest.py
seb4stien: Thanks for clarifying. We "only have one test" argument is a strong one :)

Thanhphan1147 (commented):
Looks good to me, I did a quick skim through the tests, will do a more thorough visit later

Thanhphan1147 (approved):

PR #287, opened 2025-12-12
ISD-4880 feat: grpc custom port

f-atwi:
Applicable spec: <link>

### Overview

Allows gRPC backends to select a custom frontend port.
Normally, only setting the `protocol` in the relation to https is enough to allow gRPC traffic through port 443.
If a custom port is required, the `external_grpc_port` can be passed.

### Checklist


haproxy-operator/lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -714,6 +719,49 @@ def check_services_unique(self) -> Self:
 
         return self
 
+    @model_validator(mode="after")
+    def check_external_grpc_port_unique(self) -> Self:
+        """Check that external gRPC ports are unique across units.
```
Thanhphan1147:
```suggestion
        """Check that external gRPC ports are unique across requirer applications.
```

haproxy-operator/lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -714,6 +719,49 @@ def check_services_unique(self) -> Self:
...
+        If multiple units declare the same external gRPC port,
```
Thanhphan1147:
```suggestion
        If multiple requirer applications declare the same external gRPC port,
```

haproxy-operator/lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -608,6 +610,9 @@ class RequirerApplicationData(_DatabagModel):
     allow_http: bool = Field(
         description="Whether to allow HTTP traffic in addition to HTTPS.", default=False
     )
+    external_grpc_port: int | None = Field(
+        description="Optional external gRPC port.", default=None
+    )
```
Thanhphan1147, line 615: This should be annotated to check that the value is between 0 and 65535

haproxy-operator/lib/charms/haproxy/v1/haproxy_route.py
```diff
@@ -554,11 +555,12 @@ class RequirerApplicationData(_DatabagModel):
         http_server_close: Configure server close after request.
         allow_http: Whether to allow HTTP traffic in addition to HTTPS. Defaults to False.
             Warning: enabling HTTP is a security risk, make sure you apply the necessary precautions.
+        external_grpc_port: Optional external gRPC port.
     """
 
     service: VALIDSTR = Field(description="The name of the service.")
     ports: list[int] = Field(description="The list of ports listening for this service.")
-    protocol: Optional[Literal["http", "https"]] = Field(
```
Thanhphan1147, line 561: Just a note, not sure if this will cause any side-effects to existing requirers

haproxy-operator/src/state/haproxy_route.py
```diff
@@ -178,6 +180,15 @@ def rewrite_configurations(self) -> list[str]:
...
         return rewrite_configurations
 
+    @property
+    def external_grpc_port(self) -> int | None:
+        """Get the external grpc port if grpc is used.
+
+        Returns:
+            int | None: The external grpc port.
+        """
+        return self.application_data.external_grpc_port
```
Thanhphan1147: Since we're not doing processing or adding any logic I'm not sure we should have this property as it's not adding any extra values

haproxy-operator/src/state/haproxy_route.py
```diff
@@ -320,6 +327,68 @@ def check_backend_paths(self) -> Self:
             )
         return self
 
+    @model_validator(mode="after")
+    def check_tcp_http_port_conflicts(self) -> Self:
+        """Check that TCP and HTTP backends don't use the same frontend port.
```
Thanhphan1147: Maybe we can update this docstring

haproxy-operator/src/state/haproxy_route.py
```diff
@@ -320,6 +327,68 @@ def check_backend_paths(self) -> Self:
...
+
+        Returns:
+            Self: The validated model
+        """
+        if not self.valid_backends or not self.valid_tcp_endpoints:
+            return self
```
Thanhphan1147: Not sure about this logic here, I think we want to mark each invalid relation so that we can provide feedback later in the main charm loop, but here we're skipping checking the rest of the relation data if there's already some invalid data in previous steps?

Can you maybe add a short comment explaining the logic here?

haproxy-operator/src/state/haproxy_route.py
```diff
@@ -320,6 +327,68 @@ def check_backend_paths(self) -> Self:
...
+
+        Returns:
+            list[HAProxyRouteTcpEndpoint]: List of valid TCP endpoints.
+        """
+        return [
+            endpoint
+            for endpoint in self.tcp_endpoints
+            if endpoint.relation_id not in self.relation_ids_with_invalid_data_tcp
+        ]
+
```
Thanhphan1147: since we're referencing to this multiple times maybe we can make it a `cached_property`?

haproxy-operator/templates/haproxy_route_grpc.cfg.j2
```diff
@@ -0,0 +1,23 @@
+{% set excluded_grpc_ports = [80, 443] %}
+{% for backend in backends if backend.external_grpc_port and backend.external_grpc_port not in excluded_grpc_ports and backend.servers | selectattr('protocol', 'eq', 'https') | list %}
```
Thanhphan1147: This should probably be a computed property in the charm code. I think we should try to move complex logic to the charm code and keep the templates simple.

Thanhphan1147 (commented):
First pass, a few comments. Thanks for working on this!

haproxy-operator/lib/charms/haproxy/v1/haproxy_route.py
f-atwi: I think this should not cause any side-effects.
The attribute was never optional as the field defines a default string value and should never have the value None assigned unless None is explicitly passed

haproxy-operator/src/state/haproxy_route.py
f-atwi: This bypass served a purpose but upon further inspection is not good in its correct form
In any case, I have refactored this entire method so it makes more sense

haproxy-operator/src/state/haproxy_route.py
f-atwi: I think it is best to not have them as cached properties.
it is not very expensive to generate this list and it is better to make sure that the property is up to date on each call

Thanhphan1147: The charm reinitialize itself for each hook so relation data getting out of date is probably not a concern I think, if we're calling the property only once throughout each event hook then it doesn't have a difference but if this property is accessed multiple times over the course of an event hook then we should make it a cached property, I'm fine with keeping it like this though.

f-atwi: In this model_validator here, the validator accesses this property.
https://github.com/canonical/haproxy-operator/pull/287/changes#diff-d9f89890061cfb8525413d23dc13e2755c633c07b9df15be2b251590e1fceab5R321-R371

The reason we do this is, when we are resolving port conflict between grpc and tcp, we do not want to consider backends that are invalid for because of previous validations (like invalid grpc backend because protocol is not `https`)
in the same validator, the list of invalid relation ids gets appended. Wouldn't the cached_property make the appended changes not take effect here?

docs/reference/grpc-support.md
```diff
@@ -3,8 +3,8 @@
 # gRPC support
 
 The HAProxy operator provides **automatic** gRPC support for frontends and backends using HTTPS on port 443.
+Custom gRPC frontend port can be configured using the `external_grpc_port` option in the `haproxy-route` relation.
```
arturo-seijas: A custom?

docs/reference/grpc-support.md
```diff
@@ -3,8 +3,8 @@
...
 
 ## Limitations
 
-- Only when using HTTPS
-- Only on port 443
+- Only when using HTTPS (if protocol is set to "http" in the `haproxy-route` relation, the relation will be considered invalid)
```
arturo-seijas: s/if protocol is set to "http" in the `haproxy-route` relation, the relation will be considered invalid/the relation will be considered invalid if protocol is set to "http" in the `haproxy-route` relation for simpler grammar

haproxy-operator/src/state/haproxy_route.py
arturo-seijas: As @Thanhphan1147 mentions, any time an event is fired, the charm will be instantiated. Caching the validation will only be persisted within the event handler execution

arturo-seijas (approved):
Some nitpicks. Next time, please, consider splitting into multiple PRs for easier reviewing :)

haproxy-operator/src/state/haproxy_route.py
f-atwi: but in the case of these properties, on a new config event during the validation of a model, the property will be used and the invalid_relation_ids would be updated.
wouldn't this cause the rendering of the haproxy.cfg template to be outdated as it uses the same property in the same event.
i will test locally to validate

f-atwi: tested locally with an haproxy-route and haproxy-route-tcp with conflicting port
got a conflicting haproxy configuration only with cached_property showing both routes when they should be both considered invalid

Thanhphan1147: I see, if this value changes over the course of the event hook then it should probably be a method instead of a property I think

haproxy-operator/templates/haproxy_route_grpc.cfg.j2
```diff
@@ -0,0 +1,22 @@
...
+    {% endif %}
+    use_backend {{ backend.backend_name }} if {% if backend.path_acl_required %}acl_path_{{ backend.backend_name }} {% endif %}acl_host_{{ backend.backend_name }} {% if backend.deny_path_acl_required %}!acl_deny_path_{{ backend.backend_name }}{% endif %}
+
+
+backend {{ backend.backend_name }}
+    balance {{ backend.load_balancing_configuration }}
+    timeout server {{ backend.application_data.timeout.server }}s
+    timeout connect {{ backend.application_data.timeout.connect }}s
+    {% for server in backend.servers if server.protocol == "https" %}
+    server {{ server.server_name }} {{ server.address }}:{{ server.port }} ssl ca-file {{ haproxy_cas_file }} alpn h2
```
Thanhphan1147, line 20: Note: we need to add header rewrite here to support Landscape, but that can be done in a follow-up Pr

Thanhphan1147 (approved):

docs/release-notes/artifacts/pr0287.yaml
```diff
@@ -0,0 +1,13 @@
+schema_version: 1
+changes:
+  - title: Implement custom external port for gRPC in haproxy-route relation
```
seb4stien:
```suggestion
  - title: Implemented custom external port for gRPC in haproxy-route relation
```

docs/release-notes/artifacts/pr0287.yaml
```diff
@@ -0,0 +1,13 @@
...
+    author: f-atwi
+    type: minor
+    description: |
+      Add support for custom external port for gRPC services using the config option `external_grpc_port`.
```
seb4stien:
```suggestion
      Added support for custom external port for gRPC services using the config option `external_grpc_port`.
```

haproxy-operator/lib/charms/haproxy/v1/haproxy_route.py
seb4stien: @f-atwi would it be costly to turn this into a set if we expect values to be unique?
cc @Thanhphan1147 unless there was a reason to have a list?

haproxy-operator/src/state/haproxy_route.py
```diff
@@ -320,6 +318,81 @@ def check_backend_paths(self) -> Self:
...
+                if standard_port in grpc_ports:
+                    logger.error(
+                        f"gRPC backend conflicts with HTTP backends on external port {standard_port}."
+                    )
+                    self.relation_ids_with_invalid_data.append(
+                        grpc_ports[standard_port].relation_id
+                    )
+
+        # Check for conflicts between gRPC and TCP ports
+        for port in grpc_ports.keys() & tcp_ports.keys():
```
seb4stien, line 363:
```suggestion
        for port in (grpc_ports.keys() & tcp_ports.keys()):
```

haproxy-operator/templates/haproxy_route_grpc.cfg.j2
```diff
@@ -0,0 +1,22 @@
+{% for backend in grpc_backends %}
+frontend {{ backend.backend_name }}
+    mode http
+    bind [::]:{{ backend.application_data.external_grpc_port }} v4v6 ssl crt /var/lib/haproxy/certs alpn h2
```
seb4stien, line 4: @f-atwi can you test manually on a VM that a bind[::] will work on an IPv4 only machine?
If not, please create a ticket to fix haproxy config in another PR.

haproxy-operator/templates/haproxy_route_grpc.cfg.j2
```diff
@@ -0,0 +1,22 @@
...
+    acl acl_deny_path_{{ backend.backend_name }} path_beg -i {% for path in backend.application_data.deny_paths %}{{ path }} {% endfor +%}
+    {% endif %}
+    use_backend {{ backend.backend_name }} if {% if backend.path_acl_required %}acl_path_{{ backend.backend_name }} {% endif %}acl_host_{{ backend.backend_name }} {% if backend.deny_path_acl_required %}!acl_deny_path_{{ backend.backend_name }}{% endif %}
+
+
+backend {{ backend.backend_name }}
+    balance {{ backend.load_balancing_configuration }}
+    timeout server {{ backend.application_data.timeout.server }}s
+    timeout connect {{ backend.application_data.timeout.connect }}s
+    {% for server in backend.servers if server.protocol == "https" %}
```
seb4stien: Duplicate with validator?

haproxy-operator/tests/unit/test_state.py
```diff
@@ -293,19 +230,308 @@ def test_tls_allow_no_certificate():
...
+        ca_certs_configured=True,
+    )
+
+    assert tcp_relation_id not in haproxy_route_information.relation_ids_with_invalid_data_tcp
+    assert http_relation_id not in haproxy_route_information.relation_ids_with_invalid_data
+    assert len(haproxy_route_information.valid_backends()) == 1
+    assert len(haproxy_route_information.valid_tcp_endpoints()) == 1
+
+
+def test_tcp_http_no_conflict_no_http_backends(
```
seb4stien: Simpler name maybe?
```suggestion
def test_tcp_only_happy_path(
```

haproxy-operator/tests/unit/test_state.py
```diff
@@ -293,19 +230,308 @@ def test_tls_allow_no_certificate():
...
+        peers=[],
+        ca_certs_configured=False,
+    )
+
+    assert len(haproxy_route_information.relation_ids_with_invalid_data_tcp) == 0
+    assert len(haproxy_route_information.relation_ids_with_invalid_data) == 0
+    assert len(haproxy_route_information.valid_tcp_endpoints()) == 1
+
+
+def test_tcp_http_no_conflict_no_tcp_endpoints(
```
seb4stien:
```suggestion
def test_http_only_happy_path(
```

seb4stien (approved):

haproxy-operator/src/state/haproxy_route.py
f-atwi: ruff did not like it

PR #302, opened 2025-12-18
Add the DDoS protection interface

swetha1654:
Applicable spec: <link>

### Overview

Added the `ddos_protection` interface according to the [spec](https://docs.google.com/document/d/1I_mKiCRWz6A9VIvc1moyFVNIRuJhSy8cwdGjjSPhlY0/edit?tab=t.0#heading=h.adwk3is9wch6).

### Checklist


haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
```diff
@@ -0,0 +1,474 @@
...
+    model_config = ConfigDict(
+        # tolerate additional keys in databag
+        extra="ignore",
+        # Allow instantiating this class by field name (instead of forcing alias).
+        populate_by_name=True,
+    )  # type: ignore
+    """Pydantic config."""
+
+    @classmethod
+    def load(cls, databag: MutableMapping[str, str]) -> "_DatabagModel":
```
f-atwi: nit: could be annotated to return `Self`

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
```diff
@@ -0,0 +1,474 @@
...
+        http_keepalive_timeout: Timeout for HTTP keep-alive connections in seconds.
+        client_timeout: Timeout for client connections in seconds.
+        deny_paths: List of paths to deny.
+    """
+
+    rate_limit_requests_per_minute: Optional[int] = Field(default=None, gt=0)
+    rate_limit_connections_per_minute: Optional[int] = Field(default=None, gt=0)
+    concurrent_connections_limit: Optional[int] = Field(default=None, gt=0)
+    error_rate: Optional[int] = Field(default=None, gt=0)
+    limit_policy: Optional[RateLimitPolicy] = Field(default=RateLimitPolicy.SILENT)
```
f-atwi, line 225:
```suggestion
    limit_policy: RateLimitPolicy = Field(default=RateLimitPolicy.SILENT)
```

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
```diff
@@ -0,0 +1,474 @@
...
+        try:
+            return cls.model_validate_json(json.dumps(data))
+        except ValidationError as e:
+            msg = f"failed to validate databag: {databag}"
+            logger.error(msg)
+            raise DataValidationError(msg) from e
+
+    def dump(
+        self, databag: MutableMapping[str, str] | None = None, clear: bool = True
+    ) -> MutableMapping[str, str] | None:
```
f-atwi:
```suggestion
    ) -> MutableMapping[str, str]:
```

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
```diff
@@ -0,0 +1,474 @@
...
+        """
+        if clear and databag:
+            databag.clear()
+
+        if databag is None:
+            databag = {}
+
+        dct = self.model_dump(mode="json", by_alias=True, exclude_defaults=True)
+        databag.update({k: json.dumps(v) for k, v in dct.items()})
+        return databag
```
f-atwi: Not sure if this affects the use case here, but the problem here is that this method mutates the input and returns the value. The intent is hence ambiguous whether it is called with a databag arg to expect it to be modified in place. (the docstring specifies that databag is the databag to write to which I assume that it would be mutated, but then it is also returned)
another issue is that it is not always mutated which makes the behaviour inconsistent. If `None` is passed as databag, then the value would not be mutated.

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
```diff
@@ -0,0 +1,474 @@
...
+        Extracts and stores the status code separately in policy_status_code.
+
+        Raises:
+            ValueError: When limit_policy is invalid.
+
+        Returns:
+            The validated model.
+        """
+        if (
+            not isinstance(self, dict)
```
f-atwi: wouldnt self always be a dict here?
but if it were not a dict we are returning it as is without validation?

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
```diff
@@ -0,0 +1,474 @@
...
+            raise ValueError("Status code can only be specified with 'deny' policy")
+
+        if policy == RateLimitPolicy.DENY and len(parts) > 1:
+            try:
+                status_code = int(parts[1])
+                if not (100 <= status_code <= 599):
+                    raise ValueError(
+                        f"Invalid HTTP status code '{status_code}' in limit_policy. "
+                        f"Must be between 100 and 599."
+                    )
```
f-atwi: shouldnt this be handled by pydantic automatically?

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
```diff
@@ -0,0 +1,474 @@
...
+
+        self._relation_name = relation_name
+        self.charm = charm
+
+    def _update_relation_data(self) -> None:
+        """Update the relation data with the current provider configuration."""
+        relations = self.charm.model.relations.get(self._relation_name, [])
+        for relation in relations:
+            try:
+                self._provider_data.dump(relation.data[self.charm.app], clear=True)
```
f-atwi: because of the comment I mentioned earlier
(I am not sure relation.data[self.charm.app] would ever be None, but if it is) then it would not be changed

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
```diff
@@ -0,0 +1,474 @@
...
+
+        if policy == RateLimitPolicy.DENY and len(parts) > 1:
+            try:
+                status_code = int(parts[1])
+                if not (100 <= status_code <= 599):
+                    raise ValueError(
+                        f"Invalid HTTP status code '{status_code}' in limit_policy. "
+                        f"Must be between 100 and 599."
+                    )
+                self["policy_status_code"] = status_code
```
f-atwi: what if the policy_status_code was specified explicitly too?

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
javierdelapuente: Yeah, I would not return anything if the input is mutated, following other Python functions like sorted vs list.sort.

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
```diff
@@ -0,0 +1,474 @@
...
+            ValueError: When limit_policy is invalid.
+
+        Returns:
+            The validated model.
+        """
+        if (
+            not isinstance(self, dict)
+            or "limit_policy" not in self
+            or self["limit_policy"] is None
+        ):
```
javierdelapuente: self.get("limit_policy") is None?

seb4stien (approved):

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
```diff
@@ -0,0 +1,474 @@
...
+        self.charm = charm
+
+    def _update_relation_data(self) -> None:
+        """Update the relation data with the current provider configuration."""
+        relations = self.charm.model.relations.get(self._relation_name, [])
+        for relation in relations:
+            try:
+                self._provider_data.dump(relation.data[self.charm.app], clear=True)
+            except Exception as e:
+                logger.error("Failed to update relation data: %s", str(e))
```
javierdelapuente: Why could this happen?

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
swetha1654: that makes sense. I actually copied this entire class from the haproxy_route_tcp (The same class is used in many other interfaces) so it might be good to raise another PR to update all the other libraries as well. Thanks for catching it.

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
swetha1654: since this is a "before" validator, the pydantic validation would for status codes would not have happened yet. It would only happen after this method is executed. But i guess it doesn't matter if we catch it here or there so I'll remove it.

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
swetha1654: the `set_config` method does not expose the policy_status_code so the user cannot explicitly specify it.

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
swetha1654: it would not be none since it returns a Mapping

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
swetha1654: Removing the Optional causes typing inconsistencies since the `set_config` allows limit_policy to be None

javierdelapuente (approved):

f-atwi (approved):

haproxy-operator/lib/charms/haproxy/v0/ddos_protection.py
swetha1654: indeed its not required

PR #323, opened 2026-01-14
Implement requirer side of the `ddos_protection` interface

swetha1654:
Applicable spec: <link>

### Overview

Implemented requirer side of the `ddos_protection` interface.
Split `limit_policy` into `limit_policy_http` and `limit_policy_tcp`.
Added unit and integration tests.

### Checklist


haproxy-operator/templates/haproxy_ingress.cfg.j2
```diff
@@ -23,6 +23,14 @@ frontend ingress
     default_backend default
 {% endfor %}
 
+peers haproxy_peers
```
seb4stien: Shouldn't it be in the "if" below? I don't see it in the previous template?

haproxy-operator/templates/haproxy_ingress_per_unit.cfg.j2
```diff
@@ -23,6 +24,14 @@ frontend ingress_per_unit
     default_backend default
 {% endfor %}
 
+peers haproxy_peers
```
seb4stien: Shouldn't it be in the "if" below? I don't see it in the previous template.

seb4stien (commented):
Left a question regarding how the "peers" are handled in the template if ddos is not enabled.

haproxy-operator/templates/haproxy_ingress.cfg.j2
swetha1654: good catch, I've updated the config. I'm surprised the ingress related integration tests didn't throw a template error.

seb4stien (approved):

haproxy-operator/src/state/ddos_protection.py
```diff
@@ -0,0 +1,163 @@
...
+        )
+
+        return file_path
+
+    @classmethod
+    def from_charm(
+        cls,
+        ddos_requirer: DDoSProtectionRequirer,
+    ) -> "DDosProtection":
+        """Get DDoS protection configuration from charm's ddos-protection relation.
```
javierdelapuente, line 75: The function, besides getting the DDoS protection configuration, it is writing to two files. I think these actions should be more explicit (and maybe done in the reconcile)

swetha1654: If my being more explicit you mean explicitly mentioning it in the docstring, i have updated it. WRT to doing it in the reconcile, I'm not sure if that a good idea as this is specific to DDoS only and keeping that logic in this module feels right. WDYT?

javierdelapuente: Following https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619, the state abstracts the configuration and integration data provided by Juju as well as other charm state through a model that is easier to interact with.

In this case, besides that, it is executing business logic related to writing to a configuration file. My opinion is that is not responsability of the state (as we normally define it) to write to configuration files.

swetha1654: @javierdelapuente  I don't think the ip_allow_list and deny_paths list are configuration files. They are merely the state variables stored in a file format for cleaner representation of these values in the `haproxy.cfg` rather than having messy for loops using Jinja templating. Going by the spec you sent, the charm runtime state includes "state that is stored on the filesystem of the application workload", and i think the ip allow list and deny paths list comes under that category.

seb4stien (approved):

haproxy-operator/src/state/ddos_protection.py
javierdelapuente: I do not mean the code for the list has to be in a Jinja template (the current list comprehensions are good).

What I mean, is that in the state module there is HAProxy specific configuration. I do not think in that spec the state in the filesystem refers to a file like this one, otherwise we may include in that list the haproxy.conf file.

I think that is a configuration file for HAProxy. For example, if we snap HAProxy, the locations of the configuration files may change, and they will be changed at least in the charm state module. I would not expect that change to happen there.

Same for the format of the file, if a future releare of HAProxy changes how it reads files, or if you want to list all ips in the haproxy conf file instead of pointing to a file, you will have to change both the charm state module and the haproxy module/reconcile. I would expect only the last one to change.

swetha1654: Okay, i think i understand now. What I was originally going for was trying to keep data and its representation logic close to each other for code simplicity and less duplication, but i think this makes more sense. 

However i don't think the reconcile function is the best place to move this logic to. The charm class should be used purely to interact with Juju and orchestrate responding to events from Juju. Maybe the `haproxy.py` module would be the better place. WDYT @javierdelapuente ?

javierdelapuente (approved):

haproxy-operator/src/state/ddos_protection.py
javierdelapuente: I think haproxy module as you wrote it is good!

PR #325, opened 2026-01-14
fix: add missing settings from haproxy-route-tcp relation template

skatsaounis:
Applicable spec: N/A

### Overview

The haproxy-route-tcp relation supports setting `retry_count`, `retry_redispatch`, `load_balancing_algorithm`, `load_balancing_consistent_hashing` in the `HaproxyRouteTcpRequirer` and is adding them to the relation databag with `_generate_application_data`. This PR is extending the `haproxy_route_tcp.cfg.j2` Jinja template to use them.

### Rationale

The current template is not respecting the values set to the application databag.

### Juju Events Changes

N/A

### Module Changes

N/A

### Library Changes

N/A

### Checklist

- [ ] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)

### DIFF
--- a/haproxy-operator/templates/haproxy_route_tcp.cfg.j2
+++ b/haproxy-operator/templates/haproxy_route_tcp.cfg.j2
@@ -17,6 +17,12 @@ maxconn {{ server.maxconn }}
 frontend {{ endpoint.name }}
     mode tcp
     option tcplog
+{% if endpoint.application_data.retry %}
+    retries {{ endpoint.application_data.retry.count }}
+{% if endpoint.application_data.retry.redispatch %}
+    option redispatch
+{% endif %}
+{% endif %}
     bind [::]:{{ endpoint.application_data.port }} v4v6 {% if endpoint.application_data.tls_terminate %} ssl crt {{ haproxy_crt_dir }} {% endif +%}
 {% if endpoint.application_data.enforce_tls %}
     tcp-request content accept if { req_ssl_hello_type 1 }
@@ -36,6 +42,12 @@ frontend {{ endpoint.name }}
 
 backend {{ endpoint.name }}
     mode tcp
+{% if endpoint.application_data.load_balancing %}
+    balance {{ endpoint.application_data.load_balancing.algorithm }}
+{% if endpoint.application_data.load_balancing.consistent_hashing %}
+    hash-type consistent
+{% endif %}
+{% endif %}
 {% for option in endpoint.tcp_check_options %}
     {{ option }}
 {% endfor %}
--- a/haproxy-operator/templates/haproxy_route_tcp.cfg.j2
+++ b/haproxy-operator/templates/haproxy_route_tcp.cfg.j2
@@ -43,7 +43,7 @@ frontend {{ endpoint.name }}
 backend {{ endpoint.name }}
     mode tcp
 {% if endpoint.application_data.load_balancing %}
-    balance {{ endpoint.application_data.load_balancing.algorithm }}
+    balance {{ endpoint.application_data.load_balancing.algorithm.value }}
 {% if endpoint.application_data.load_balancing.consistent_hashing %}
     hash-type consistent
 {% endif %}
--- a/haproxy-operator/tests/integration/conftest.py
+++ b/haproxy-operator/tests/integration/conftest.py
@@ -307,7 +307,7 @@ def any_charm_haproxy_route_requirer_fixture(
     )
 
 
-@pytest.fixture(scope="function", name="any_charm_haproxy_route_tcp_requirer_base")
+@pytest.fixture(scope="module", name="any_charm_haproxy_route_tcp_requirer_base")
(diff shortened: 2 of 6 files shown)


haproxy-operator/templates/haproxy_route_tcp.cfg.j2
```diff
@@ -36,6 +42,12 @@ frontend {{ endpoint.name }}
 
 backend {{ endpoint.name }}
     mode tcp
+{% if endpoint.application_data.load_balancing %}
+    balance {{ endpoint.application_data.load_balancing.algorithm }}
```
javierdelapuente: I think there is a bug here, as this will write something like: LoadBalancingAlgorithm.LEASTCONN instead of leastconn. I think a test could be useful to find this kind of issues.

javierdelapuente (commented):
Thank you very much for the contribution. I’ve left a comment — feel free to ping me if needed.

haproxy-operator/tests/integration/conftest.py
```diff
@@ -307,14 +307,19 @@ def any_charm_haproxy_route_requirer_fixture(
...
+@pytest.fixture(scope="function", name="any_charm_haproxy_route_tcp_requirer_base")
 @pytestconfig_arg_no_deploy(application=ANY_CHARM_HAPROXY_ROUTE_TCP_REQUIRER_APPLICATION)
-def any_charm_haproxy_route_tcp_requirer_fixture(
+def any_charm_haproxy_route_tcp_requirer_base_fixture(
     _pytestconfig: pytest.Config, juju: jubilant.Juju
 ):
     """Deploy any-charm and configure it to serve as a requirer for the haproxy-route
     integration.
     """
+    if ANY_CHARM_HAPROXY_ROUTE_TCP_REQUIRER_APPLICATION in juju.status().apps:
```
skatsaounis: I am wondering why I had to do this. Maube I have to change scope function to scope module at line 310. Please assist @javierdelapuente

haproxy-operator/tests/integration/test_haproxy_route_tcp.py
```diff
@@ -13,7 +13,7 @@
 
 
 @pytest.mark.abort_on_fail
-async def test_haproxy_route_tcp(
+def test_haproxy_route_tcp(
```
skatsaounis, line 16: I removed async so that to match the test cases of haproxy_route_requirer.py. If it's not ok let me know to revert @javierdelapuente

haproxy-operator/templates/haproxy_route_tcp.cfg.j2
skatsaounis: you were right. I had to add `.value` and the test revealed that.

haproxy-operator/tests/integration/conftest.py
javierdelapuente: as we are returning, the app will not be removed even if it is function scope. If we want to have it function scope, I would yield and then remove the app in this fixture.

As you say, if you do not want to remove it, just scope module is enough instead of this check

haproxy-operator/tests/integration/conftest.py
```diff
@@ -340,3 +345,32 @@ def any_charm_haproxy_route_tcp_requirer_fixture(
...
+
+
+@pytest.fixture(name="any_charm_haproxy_route_tcp_requirer")
+def any_charm_haproxy_route_tcp_requirer_fixture(
+    any_charm_haproxy_route_tcp_requirer_base: str,
+    juju: jubilant.Juju,
+):
+    """Provide haproxy route tcp requirer and clean up test-specific relations after each test.
+
+    This function-scoped fixture wraps the module-scoped any_charm_haproxy_route_tcp_requirer_base
```
javierdelapuente, line 352: I think we could integrate haproxy and any_charm_haproxy_route_tcp_requirer_base in here, and then remove that relation at the end. That way the test will be very simple (and we know which relation to remove in this fixture.

haproxy-operator/tests/integration/conftest.py
skatsaounis: Thanks for the explanation

haproxy-operator/tests/integration/conftest.py
javierdelapuente: I close this comment. 

As talked in a message, I rather prefer test to create and clean themselves, but let's tackle this in a future pr.

skatsaounis: I intentionally mimicked the haproxy_route tests and fixtures for two reasons: 1) to be easily extended the next time someone tries to add a new test. If that someone is not part of the maintainers might spend a lot of time like me trying to understand what's going on 2) I am a fan of following the same behavior of similar code in a code base. If you observe the haproxy_route test cases, all of the integrate as part of the test case instead of a fixture.

Is it ok if it will be changed properly by the maintainers in a future PR that improves the cleanup of all fixtures, including the `any_charm_haproxy_route_tcp_requirer_fixture`?

javierdelapuente (approved):
Thanks for the PR

alithethird (approved):

PR #355, opened 2026-01-29
update state logic to merge backends

### LINKED ISSUE #352: TCP frontends are not merged together when listening on the same port
### Enhancement Proposal

Currently, we're creating one frontend section per `haproxy-route-tcp` relation.

```
{% for endpoint in tcp_endpoints %}
frontend {{ endpoint.name }}
    ...
    bind [::]:{{ endpoint.application_data.port }} v4v6 {% if endpoint.application_data.tls_terminate %} ssl crt {{ haproxy_crt_dir }} {% endif +%}
```

This is a problem when 2 requirer applications request for the same port but with different sni ( which is supposed to work )

```
juju deploy ingress-configurator conf1 --config tcp-hostname=example.com --config tcp-frontend-port=443 ...
juju deploy ingress-configurator conf2 --config tcp-hostname=test.example.com --config tcp-frontend-port=443 ...
juju relate conf1 haproxy
juju relate conf2 haproxy
```

The following deployment will create 2 `frontend` sections in the HAProxy config, both with `bind ::443`, haproxy should merge the 2 frontends into a single one with the appropriate ACLs

Thanhphan1147:
Closes #352 

### Overview

This PR rewrites the state management logic for `haproxy_route_tcp` to merge backends requesting the same port. The general logic used for merging backends that request for the same port follow these 3 rules:

```
R1: If only plain TCP is requested on the same port, there should only be one backend, any other configurations will be considered invalid
R2: If  there are more than one backends, `enforce_tls` and `sni` must be set, the backends that don't meet this requirement will be considered invalid
R3: For backends that already set `enforce_tls` and `sni`, if there are any backend that sets `tls_terminate=True`, only these backends will be used, the rest will be considered invalid
```

The PR also reworks how the haproxy_route_tcp template is rendered, emphasizing more on generating the values inside the charm code itself to reduce the complexity of the jinja2 template.

Note that `ip_deny_list` attribute is temporary ignored in the template because of potentially overlapping with ddos protection's ip_allow_list. We can also think about deprecating this attribute in the future.
 
### Checklist


haproxy-operator/src/state/haproxy_route.py
```diff
@@ -276,14 +282,17 @@ def from_provider(  # pylint: disable=too-many-arguments
...
-            relation_ids_with_invalid_data_tcp = tcp_requirers.relation_ids_with_invalid_data
-            tcp_endpoints.extend(
-                HAProxyRouteTcpEndpoint.from_haproxy_route_tcp_requirer_data(tcp_requirer)
-                for tcp_requirer in tcp_requirers.requirers_data
+            tcp_frontends = parse_haproxy_route_tcp_requirers_data(tcp_requirers)
+            # Calculate the invalid relation ids after parsing the relations data into
+            # HAProxyRouteTcpFrontend objects
+            relation_ids_with_invalid_data_tcp = (
+                tcp_requirers.relation_ids_with_invalid_data.union(
+                    *[frontend.relation_ids_with_invalid_data for frontend in tcp_frontends]
```
seb4stien, line 295:
```suggestion
                    frontend.relation_ids_with_invalid_data for frontend in tcp_frontends
```

seb4stien: To be tested, but I think it should do the job

haproxy-operator/src/state/haproxy_route.py
```diff
@@ -486,3 +501,29 @@ def generate_hostname_acls(
...
+
+def parse_haproxy_route_tcp_requirers_data(
+    tcp_requirers: HaproxyRouteTcpRequirersData,
+) -> list[HAProxyRouteTcpFrontend]:
+    """Parse HAProxyRouteTcpFrontend data from requirers into frontend objects.
+
+    Returns:
+        list[HAProxyRouteTcpFrontend]: The parsed frontend data.
+    """
+    port_to_backends_mapping: dict[int, list[HAProxyRouteTcpEndpoint]] = {}
```
seb4stien: Maybe using a defaultdict?

haproxy-operator/src/state/haproxy_route_tcp.py
```diff
@@ -140,3 +158,157 @@ def tcp_check_options(self) -> list[str]:
...
+        if len(backends) == 1:
+            return cls(
+                port=backends[0].application_data.port,
+                backends=backends,
+                enforce_tls=backends[0].application_data.enforce_tls,
+                tls_terminate=backends[0].application_data.tls_terminate,
+                relation_ids_with_invalid_data=set[int](),
+            )
+
+        # At this point we have more than one backend, all of them need to set enforce_tls=True
```
seb4stien, line 205: I may be something, or it may be worth having an additional comment here: do all backends have the same port when we call this function? If not, SNI is only required on the one on the same port, no?

seb4stien:
It may be good to document the 3 rules in the code or in an ADR.
I would need some clarifications on some of them:

> R1: If only TCP is requested on the same port, there should only be one backend, configuration will be considered invalid

I'm not sure about what you mean here. "same port" as http? Is it saying that TCP cannot be enabled on a port used by http/https?

> R2: If  there are more than one backends, `enforce_tls` and `sni` must be set, the relations that don't meet this requirement will be considered invalid

"more than one backends" for a given port?

> R3: For backends that already set `enforce_tls` and `sni`, if there are any backend that sets `tls_terminate=True`, only these backends will be used, the rest will be considered invalid

For a given port?

haproxy-operator/src/state/haproxy_route_tcp.py
```diff
@@ -140,3 +158,157 @@ def tcp_check_options(self) -> list[str]:
...
+
+    Attrs:
+        port: The port exposed on the provider.
+        backends: List of backend endpoints for this frontend.
+        enforce_tls: Whether to enforce TLS for all traffic.
+        tls_terminate: Whether to enable TLS termination.
+    """
+
+    port: int = Field(description="The port exposed on the provider.", gt=0, le=65535)
+    backends: list[HAProxyRouteTcpEndpoint] = Field(description="List of backend endpoints.")
```
Thanhphan1147: rename this to HAProxyRouteTcpBackend

haproxy-operator/src/state/haproxy_route_tcp.py
```diff
@@ -140,3 +158,157 @@ def tcp_check_options(self) -> list[str]:
...
+    enforce_tls: bool = Field(description="Whether to enforce TLS for all traffic.", default=True)
+    tls_terminate: bool = Field(description="Whether to enable tls termination.", default=True)
+    relation_ids_with_invalid_data: set[int] = Field(
+        description="List of relation ids with invalid data.", default=set[int]()
+    )
+
+    @classmethod
+    def from_backends(cls, backends: list[HAProxyRouteTcpEndpoint]) -> "Self":
```
Thanhphan1147: see if it's better to instantiate first and then validate.

haproxy-operator/src/state/haproxy_route_tcp.py
```diff
@@ -140,3 +158,157 @@ def tcp_check_options(self) -> list[str]:
...
+            backends: List of backend endpoints.
+
+        Raises:
+            HAProxyRouteTcpFrontendValidationError: When the frontend is initialized with no backends.
+
+        Returns:
+            Self: The instantiated HAProxyRouteTcpFrontend class.
+        """
+        # If there's only one backend, return the class directly with values from the backend
+        if len(backends) == 1:
```
Thanhphan1147, line 196: see if we can remove this since the logic should already be covered in the rest of the method

haproxy-operator/src/state/haproxy_route.py
```diff
@@ -486,3 +501,29 @@ def generate_hostname_acls(
...
+    for requirer in tcp_requirers.requirers_data:
+        endpoint = HAProxyRouteTcpEndpoint.from_haproxy_route_tcp_requirer_data(requirer)
+        port = endpoint.application_data.port
+        if port not in port_to_backends_mapping:
+            port_to_backends_mapping[port] = []
+        port_to_backends_mapping[port].append(endpoint)
+    tcp_frontends: list[HAProxyRouteTcpFrontend] = []
+    for backends in port_to_backends_mapping.values():
+        try:
+            frontend = HAProxyRouteTcpFrontend.from_backends(backends)
```
arturo-seijas, line 522: I'm not sure I get this. Is the frontend the same as the backend minus the port? If so,  if feels more readable to use a comprehension list.
Also, if you are only returning the `tcp_frontends`, why fetching the port from the backend, storing that into a dict with it backend...?

haproxy-operator/src/state/haproxy_route_tcp.py
```diff
@@ -140,3 +158,157 @@ def tcp_check_options(self) -> list[str]:
...
+                if backend.application_data.tls_terminate:
+                    routable_backends_with_tls_terminate.append(backend)
+                else:
+                    routable_backends_without_tls_terminate.append(backend)
+            else:
+                relation_ids_with_invalid_data.add(backend.relation_id)
+
+        if routable_backends_with_tls_terminate:
+            relation_ids_with_invalid_data.update(
+                backend.relation_id for backend in routable_backends_without_tls_terminate
```
arturo-seijas, line 225: Why making `routable_backends_without_tls_terminate` as invalid but not `routable_backends` if `routable_backends_with_tls_terminate` is non-empty?
Please, make sure this is unit tested

haproxy-operator/src/state/haproxy_route_tcp.py
```diff
@@ -140,3 +158,157 @@ def tcp_check_options(self) -> list[str]:
...
+            )
+
+        rendered_backends = routable_backends_with_tls_terminate or routable_backends
```
arturo-seijas: Didn't you forget here about `routable_backends_without_tls_terminate`? those are never returned even if there are no more valid backends.
Please, make sure this is unit tested

haproxy-operator/src/state/haproxy_route_tcp.py
```diff
@@ -140,3 +158,157 @@ def tcp_check_options(self) -> list[str]:
...
+    @property
+    def default_backend_name(self) -> str:
+        """Get the default backend name for this frontend.
+
+        The default backend is used when no SNI ACLs match.
+
+        Returns:
+            str: The name of the default backend.
+        """
+        return f"haproxy_route_tcp_{self.port}_default_backend"
```
Thanhphan1147, line 289: Add the plain TCP backend to the default backend as well

haproxy-operator/templates/haproxy_route_tcp.cfg.j2
```diff
@@ -45,31 +37,56 @@ frontend {{ endpoint.name }}
     http-request allow if allowed_ips
 {% endif %}
 
-{% if endpoint.application_data.sni %}
-    acl sni_{{ endpoint.application_data.port }} {% if endpoint.application_data.tls_terminate %}ssl_fc_sni{% else %}req_ssl_sni{%endif%} -m str {{ endpoint.application_data.sni }}
-    tcp-request content reject if !sni_{{ endpoint.application_data.port }}
+{# This configuration block is only applicable when not terminating TLS. #}
+{% if not frontend.tls_terminate %}
+{% if frontend.content_inspect_delay_required %}
```
Thanhphan1147, line 42: be more consistent with what is explicit in the j2 template and what's computed in the classes

haproxy-operator/src/state/haproxy_route_tcp.py
Thanhphan1147: yes, the list of backends passed to this method should all share the same port

haproxy-operator/src/state/haproxy_route_tcp.py
Thanhphan1147: This is intended, and this is implementing rule 3:
> R3: For backends that already set `enforce_tls` and `sni`, if there are any backend that sets `tls_terminate=True`, only these backends will be used, the rest will be considered invalid

`routable_backends_with_tls_terminate` is a subset of `routable_backends` which sets `tls_terminate=True`, so only these backend will be rendered **if the list is not empty**. If `routable_backends_with_tls_terminate` is empty then `routable_backends_without_tls_terminate` is the same as `routable_backends` so whatever we choose to render will be the same

haproxy-operator/src/state/haproxy_route_tcp.py
Thanhphan1147: This is similar to my reply below about rule 3:
> R3: For backends that already set enforce_tls and sni, if there are any backend that sets tls_terminate=True, only these backends will be used, the rest will be considered invalid

if `routable_backends_with_tls_terminate` is not empty, then anything that doesn't set `tls_terminate=True` is considered invalid. `routable_backends` is the full list of backends that met the condition of rule 2:

> R2: If  there are more than one backends, `enforce_tls` and `sni` must be set, the relations that don't meet this requirement will be considered invalid

So in essence `routable_backends` = `routable_backends_with_tls_terminate` + `routable_backends_without_tls_terminate`

haproxy-operator/src/state/haproxy_route_tcp.py
```diff
@@ -140,3 +158,157 @@ def tcp_check_options(self) -> list[str]:
...
+                port=backends[0].application_data.port,
+                backends=backends,
+                enforce_tls=backends[0].application_data.enforce_tls,
+                tls_terminate=backends[0].application_data.tls_terminate,
+                relation_ids_with_invalid_data=set[int](),
+            )
+
+        # At this point we have more than one backend, all of them need to set enforce_tls=True
+        # and have an sni value for them to be routable and merged.
+        routable_backends: list[HAProxyRouteTcpEndpoint] = []
```
javierdelapuente: If HAProxyRouteTcpEndpoint is hashable  (I think relation id identifies it) and order is not important, playing with sets could make the code simpler, something like:
```
valid_backends = set()
invalid_backends = set()

all_backends = set(backends) 

tls_backends = {b for b in all_backends if b.tls and b.sni}
sni_backends = {b for b in all_backends if not b.tls and b.enforce_tls and b.sni} 
other_backends = all_backends - tls_backends - sni_backends

if tls_backends:
    valid_backends = tls_backends
    invalid_backends = sni_backends | other_backends
elif sni_backends:
    valid_backends = sni_backends
    invalid_backends = other_backends
else:
    if len(other_backends) > 1:
        invalid_backends = other_backends
    else:
        valid_backends = other_backends
```

haproxy-operator/tests/unit/test_state.py
```diff
@@ -711,3 +718,498 @@ def test_ddos_protection_from_charm_relation_data_error():
 
     assert "Failed to load DDoS protection configuration" in str(exc_info.value)
     assert "Invalid relation data" in str(exc_info.value)
+
+
+def test_haproxy_route_tcp_frontend_from_backends_single_backend(
```
javierdelapuente, line 723: I think it is difficult to follow the tests. Maybe have some fixtures/constants with backends so they are not repeated in many tests and then tests are shorter and easier to follow?

Thanhphan1147:
> R1: If only TCP is requested on the same port, there should only be one backend, configuration will be considered invalid

I'm not sure about what you mean here. "same port" as http? Is it saying that TCP cannot be enabled on a port used by http/https?

-> I'll update the wording here to "plain TCP", here I'm making the distinction between plain TCP and TLS.

> R2: If  there are more than one backends, `enforce_tls` and `sni` must be set, the relations that don't meet this requirement will be considered invalid

"more than one backends" for a given port?

-> Yes, here we're talking about merging backends that request for the same port under a single frontend

> R3: For backends that already set `enforce_tls` and `sni`, if there are any backend that sets `tls_terminate=True`, only these backends will be used, the rest will be considered invalid

For a given port?

-> Yes, here we're talking about merging backends that request for the same port under a single frontend

haproxy-operator/src/state/haproxy_route.py
Thanhphan1147: Not sure what is the advantage of a default dict here since we don't access this dict by key, rather we loop through the list of values and create a frontend for each one:

```
    for backends in port_to_backends_mapping.values():
        try:
            frontend = HAProxyRouteTcpFrontend.from_backends(backends)
```

Can you elaborate further on this?

haproxy-operator/src/state/haproxy_route.py
Thanhphan1147: didn't seem to work so I reverted the commit:
```
FAILED tests/unit/test_state.py::test_haproxy_route_requirer_information_reserved_ports - TypeError: unhashable type: 'set'
```

haproxy-operator/src/state/haproxy_route_tcp.py
Thanhphan1147: done

haproxy-operator/src/state/haproxy_route.py
Thanhphan1147: Here we have `tcp_requirers.requirers_data` which is a list of backends from all of the `haproxy-route-tcp` requirers and in there we can have backends that request for the same frontend port ( e.g 2 backends that request for frontend port 443 ). We should merge them together so that's why we're grouping the backends by port into a dictionary.

So in general 4 backends might not necessarily give 4 frontends, because backends that request the same port will be merged together into a single frontend.

Let me know if that answers your question!

haproxy-operator/src/state/haproxy_route_tcp.py
Thanhphan1147: The way the logic is structured means that we can't remove this

haproxy-operator/src/state/haproxy_route_tcp.py
Thanhphan1147: done

haproxy-operator/src/state/haproxy_route_tcp.py
arturo-seijas: Gotcha. Can't you simplify this logic by setting `relation_ids_with_invalid_data = backends - rendered_backends` on line 330 and remove lines 221 - 226 or am I missing something here? Also, you don't need `routable_backends`, you can set `rendered_backends = routable_backends_with_tls_terminate or routable_backends_without_tls_terminate`, which is equivalent

haproxy-operator/src/state/haproxy_route.py
```diff
@@ -486,3 +501,29 @@ def generate_hostname_acls(
...
+
+def parse_haproxy_route_tcp_requirers_data(
+    tcp_requirers: HaproxyRouteTcpRequirersData,
+) -> list[HAProxyRouteTcpFrontend]:
+    """Parse HAProxyRouteTcpFrontend data from requirers into frontend objects.
+
+    Returns:
+        list[HAProxyRouteTcpFrontend]: The parsed frontend data.
+    """
+    port_to_backends_mapping: dict[int, list[HAProxyRouteTcpBackend]] = {}
```
arturo-seijas: Maybe a defaultdict would make the code a bit simpler

haproxy-operator/src/state/haproxy_route.py
arturo-seijas: It does. Thank you for clarifying

haproxy-operator/src/state/haproxy_route.py
Thanhphan1147: updated this to a defaultdict

haproxy-operator/src/state/haproxy_route.py
Thanhphan1147: thanks, updated this to a defaultdict

seb4stien (approved):
Please ensure conflicts are tested.

haproxy-operator/src/state/haproxy_route_tcp.py
Thanhphan1147: I think to create a set of HAProxyRouteTcpBackend we'd need to add implementation for `__hash__` or `__eq__`, which would simplify the logic indeed but I'd like to do it in a follow-up PR if that's possible

haproxy-operator/src/state/haproxy_route_tcp.py
Thanhphan1147: That's a good point, I removed the `routable_backends` and updated the code. For simplifying the logic I think Javi also made a good point about managing the backends as sets instead of lists, which would also help simplifying the logic here, but I'll probably do it in a follow-up PR.

haproxy-operator/tests/unit/test_state.py
Thanhphan1147: `haproxy_route_tcp_relation_data` is already a fixture that generates the required class, I actually prefers this way because it clearly shows what's the input and expected result for each tests in a vacuum without having to go search for a fixture that abstracted away the input. Is there a test in particular that you find too long or difficult to follow?

javierdelapuente: I see many repeated "data preparation"  steps in all tests (basically they are all the same). I was just looking at the names to understand them. But I think your point is fair, they are not so long.

javierdelapuente (approved):

haproxy-operator/src/state/haproxy_route_tcp.py
Thanhphan1147: will evaluate in a follow-up PR to refactor

haproxy-operator/templates/haproxy_route_tcp.cfg.j2
Thanhphan1147: will evaluate in a follow-up PR

arturo-seijas (approved):

PR #376, opened 2026-02-16
Bootstrap copilot collections

seb4stien:
Onboard copilot collections

### DIFF
--- a/.github/workflows/copilot-collections-update.yml
+++ b/.github/workflows/copilot-collections-update.yml
@@ -1,7 +1,7 @@
 name: Auto-Update Copilot Instructions
 on:
   schedule:
-    - cron: '0 9 * * 1' # Run every Monday at 09:00 UTC
+    - cron: '3 3 * * 1'
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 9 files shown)


swetha1654 (commented):
Does GH copilot provide a way to source the copilot instructions from an upstream repository? If these instructions become outdated it would be very difficult to maintain them across all our repos.

seb4stien:
Oh. You're right, I forgot to add the auto-update action :facepalm:

arturo-seijas (approved):

seb4stien:
@swetha1654 I added the auto-update workflow.

.github/workflows/copilot-collections-update.yml
```diff
@@ -0,0 +1,10 @@
+name: Auto-Update Copilot Instructions
+on:
+  schedule:
+    - cron: '0 9 * * 1' # Run every Monday at 09:00 UTC
```
swetha1654: i'm not sure if we have the same type of PRs across other repos, but if we do we should probably consider spreading the schedule a bit to avoid burdening the self hosted runners.

swetha1654 (approved):

.github/workflows/copilot-collections-update.yml
seb4stien: Good point. I changed the schedule. Thanks.

PR #413, opened 2026-03-26
Haproxy route policy switch default db to postgresql

Thanhphan1147:
Switch the default db engine of the policy app to postgresql.
Added psycopg-binary as a dependency

### Checklist


arturo-seijas (approved):

haproxy-route-policy/haproxy_route_policy/settings.py
```diff
@@ -76,9 +77,16 @@
 
 DATABASES = {
     "default": {
-        "ENGINE": "django.db.backends.sqlite3",
-        "NAME": BASE_DIR / "db.sqlite3",
-    }
+        "ENGINE": "django.db.backends.postgresql",
+        "PASSWORD": os.getenv("DJANGO_DATABASE_PASSWORD", ""),
```
seb4stien: Shouldn't it refuste to start without a password? What are the consequences of having PASSWORD as an empty string?

seb4stien (commented):
Not sure about the default value for password.

haproxy-route-policy/haproxy_route_policy/settings.py
Thanhphan1147: I guess this would be useful in testing if we allow connecting to the DB without a password

Thanhphan1147: If we have a DB connection issue it'll be wrapped in a 500 HTTP error message using the middleware that we defined

seb4stien: No db should be configured with no password, even in dev ;)
If we want to have a default value here, I'd use a generic one: "postgres" for instance.

seb4stien (approved):

haproxy-route-policy/haproxy_route_policy/settings.py
Thanhphan1147: Makes sense, I updated the default password to "postgres"

PR #422, opened 2026-03-31
[haproxy-route-policy-operator] add secret handling, update tests

Thanhphan1147:
### Checklist


haproxy-route-policy-operator/src/charm.py
```diff
@@ -54,15 +73,31 @@ def __init__(self, *args: Any):
 
     def _reconcile(self, _: ops.EventBase) -> None:
         """Reconcile snap configuration and service state."""
+        peer_relation = self.model.get_relation(PEER_RELATION_NAME)
+        if not peer_relation:
+            self.unit.status = ops.WaitingStatus("Waiting for peer relation.")
```
seb4stien, line 78: Will we remain in waiting if there's only one unit? Or is there a peer-relation in this situation too?

haproxy-route-policy-operator/src/charm.py
```diff
@@ -54,15 +73,31 @@ def __init__(self, *args: Any):
...
             database_information = DatabaseInformation.from_requirer(self, self.database)
-            configure_snap(database_information.haproxy_route_policy_snap_configuration)
+            configure_snap(
+                {
+                    **self._get_django_secret_key(peer_relation),
+                    **database_information.haproxy_route_policy_snap_configuration,
+                }
+            )
             self.unit.status = ops.MaintenanceStatus("running database migrations")
             run_migrations()
```
seb4stien: I just realised the migrations should only be run on the leader, is that the case?

haproxy-route-policy-operator/src/charm.py
```diff
@@ -75,9 +110,82 @@ def _reconcile(self, _: ops.EventBase) -> None:
...
+        """Get the Django secret key from the charm's config.
+
+        Returns:
+            The Django secret key.
+
+        Raises:
+            DjangoSecretKeyMissingError: If the secret key is not yet created by the leader.
+        """
+        try:
+            secret = self.model.get_secret(label=DJANGO_SECRET_KEY_SECRET_LABEL)
```
seb4stien, line 135: Isn't there a native way for django to store it in the database so that we don't have to implement the logic in the charm?

haproxy-route-policy-operator/src/charm.py
```diff
@@ -75,9 +110,82 @@ def _reconcile(self, _: ops.EventBase) -> None:
...
+
+        Returns:
+            The Django admin user.
+        """
+        try:
+            secret = self.model.get_secret(label=DJANGO_ADMIN_CREDENTIALS_SECRET_LABEL)
+            return secret.get_content()
+        except ops.SecretNotFoundError:
+            if self.unit.is_leader():
+                django_admin_credentials_data = {
```
seb4stien, line 162: Same as above: is it necessarilty something we have to do in the charm? Cannot django rely on the database?

seb4stien (commented):
Left some comments.

haproxy-route-policy-operator/src/charm.py
Thanhphan1147: There should still be a peer relation even if it's one unit, so I think we're covered

haproxy-route-policy-operator/src/charm.py
Thanhphan1147: I updated it to run on the leader

haproxy-route-policy-operator/src/charm.py
Thanhphan1147: We can't store the secret key in DB because it's read during startup, and to connect to the DB the secret-key is required so storing it in DB would create a circular dependency iiuc

haproxy-route-policy-operator/src/charm.py
Thanhphan1147: I think the problem here is that the credentials are required to be able to call the API and the DB is behind the API so having a fully-exposed API endpoint to fetch the admin credentials is a bit of an anti-pattern and I guess wouldn't be too good in terms of security as well.

haproxy-route-policy-operator/src/charm.py
seb4stien: Ok, that's the comment I lost. So it was there somewherre :)

haproxy-route-policy-operator/charmcraft.yaml
```diff
@@ -31,7 +31,7 @@ links:
     - https://launchpad.net/~canonical-is-devops
 
 assumes:
-  - juju >= 3.6
+  - juju >= 3.3
```
seb4stien: Why switching back to 3.3?

seb4stien (approved):

haproxy-route-policy-operator/tests/unit/test_charm.py
```diff
@@ -62,3 +82,69 @@ def test_config_changed_reconciles_snap_with_database_credentials():
...
+    with (
+        patch("charm.install_snap"),
+        patch("charm.configure_snap"),
+        patch("charm.run_migrations"),
+        patch("charm.start_gunicorn_service"),
+        patch("charm.create_or_update_user"),
+    ):
+        out = ctx.run(ctx.on.config_changed(), state)
+
+    assert len(list(out.secrets)) == 2
```
javierdelapuente, line 158: Should we test any unit/app status in here? Also, could we test for the specific secrets?

haproxy-route-policy-operator/src/charm.py
```diff
@@ -75,9 +112,82 @@ def _reconcile(self, _: ops.EventBase) -> None:
...
+        try:
+            secret = self.model.get_secret(label=DJANGO_SECRET_KEY_SECRET_LABEL)
+            return secret.get_content()
+        except ops.SecretNotFoundError:
+            if self.unit.is_leader():
+                django_secret_key_data = {"secret-key": secrets.token_urlsafe(32)}
+                secret = self.app.add_secret(
+                    label=DJANGO_SECRET_KEY_SECRET_LABEL, content=django_secret_key_data
+                )
+                peer_relation.data[self.app]["django-secret-key"] = (
```
javierdelapuente, line 143: Why do we need to put the label in the peer relation? I mean, it is a constant that all the units know.

Thanhphan1147: Yes the label is unnecessary but we need to put something there to trigger the `peer-relation-changed` event so that the units can reconcile and fetch the secrets. I decided to just put the label here.

haproxy-route-policy-operator/tests/unit/test_charm.py
```diff
@@ -47,13 +54,26 @@ def test_config_changed_reconciles_snap_with_database_credentials():
...
+            testing.Secret(
+                label=DJANGO_SECRET_KEY_SECRET_LABEL, tracked_content={"secret-key": "test"}
+            ),
+            testing.Secret(
+                label=DJANGO_ADMIN_CREDENTIALS_SECRET_LABEL,
+                # Ignore bandit warning as this is for testing.
+                tracked_content={"username": "admin", "password": "admin"},  # nosec
+            ),
+        ],
+    )
```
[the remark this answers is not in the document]
Thanhphan1147: fixed

haproxy-route-policy-operator/charmcraft.yaml
Thanhphan1147: updated. This was due to some issues with rebasing

haproxy-route-policy/policy/management/commands/upsertsuperuser.py
```diff
@@ -0,0 +1,34 @@
...
+            password = os.environ["DJANGO_SUPERUSER_PASSWORD"]
+        except KeyError:
+            raise CommandError("environment variable DJANGO_SUPERUSER_PASSWORD not set")
+        email = os.environ.get("DJANGO_SUPERUSER_EMAIL", "")
+        user, created = User.objects.get_or_create(username=username)
+        if created or not user.check_password(password):
+            user.set_password(password)
+        user.is_staff = True
+        user.is_superuser = True
+        user.email = email
```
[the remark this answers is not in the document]
Thanhphan1147: This user is supposed to only be managed by the charm, so this should be fine

haproxy-route-policy-operator/tests/unit/test_charm.py
Thanhphan1147: The charm creates a random one so I don't think we need to test for the secret values, I can add a test for the expected length ( 43 ) and add also a status check

haproxy-route-policy-operator/src/charm.py
javierdelapuente: don't they get the https://documentation.ubuntu.com/juju/3.6/reference/hook/#secret-changed event?

Thanhphan1147: I didn't know about this event, let me do some quick tests

Thanhphan1147: It seems that the secrets-related events are designed for the secret owner and consumer to be on different sides of a relation ( https://documentation.ubuntu.com/ops/latest/howto/manage-secrets/ ) And from my testing locally I couldn't get the event to fire to align the units. So I'll investigate on this further but for now I'll keep the PR as it is I think

PR #423, opened 2026-04-02
[haproxy-route-policy-operator] move juju secret handling to charm state, add charm state for policy related values, add unit tests.

Thanhphan1147:
**Charm state management and configuration:**
* Introduced `state/policy.py` containing the `HaproxyRoutePolicyInformation` dataclass, error classes, and logic for managing Django secrets and allowed hosts, including validation of FQDNs and IP addresses
* Added the `allowed-hosts` configuration option to `charmcraft.yaml`, enabling users to specify allowed host/domain names for the API, which are validated and passed to the application

**Charm code refactor and improvements:**
* Refactored `charm.py` to use the new state management module, removing duplicate logic and error classes, and updating reconciliation to use the new dataclass for configuration and secrets
* Improved error handling in `_reconcile` to properly manage peer relation and secret readiness, using new custom exceptions

**Dependency and validation enhancements:**
* Added the `validators` library as a dependency for domain validation and ensured all hostnames in `allowed-hosts` are validated
**Testing improvements:**
* Added a new test suite for `HaproxyRoutePolicyInformation`, covering allowed hosts validation, required fields, and snap configuration serialization.
* Updated charm unit tests to cover peer relation handling and secret creation for both leader and non-leader units, and to use the new configuration and state logic

### Checklist


haproxy-route-policy-operator/charmcraft.yaml
```diff
@@ -58,3 +58,12 @@ charm-libs:
...
     interface: haproxy_route_policy_peer
+
+config:
+  options:
+    allowed-hosts:
+      type: string
+      description: A comma-separated list of host/domain names that the dns-policy-app API
+        can serve. This configuration will set the DJANGO_ALLOWED_HOSTS environment
+        variable with its content being a JSON encoded list.
+      default: "0.0.0.0"
```
alithethird:
```suggestion
      default: "0.0.0.0"

```

alithethird (approved):
1 nitpick otherwise LGTM!

haproxy-route-policy-operator/src/state/policy.py
```diff
@@ -0,0 +1,164 @@
...
+
+        Returns:
+            HaproxyRoutePolicyInformation: The information.
+
+        Raises:
+            PeerRelationMissingError: If the peer relation is missing.
+        """
+        peer_relation = charm.model.get_relation(PEER_RELATION_NAME)
+        if not peer_relation:
+            raise PeerRelationMissingError("Peer relation is missing.")
```
seb4stien, line 95: What would happen if there's only a single unit? Couldn't it result in blocking the charm?

haproxy-route-policy-operator/src/charm.py
```diff
@@ -94,84 +85,37 @@ def _reconcile(self, _: ops.EventBase) -> None:
...
         except DatabaseRelationMissingError:
             self.unit.status = ops.BlockedStatus("Missing database relation.")
             return
         except DatabaseRelationNotReadyError:
             logger.exception("Database relation not ready")
             self.unit.status = ops.WaitingStatus("waiting for complete database relation.")
             return
+        except PeerRelationMissingError:
+            logger.exception("Peer relation missing")
+            self.unit.status = ops.WaitingStatus("Waiting for peer relation.")
```
seb4stien, line 114: Related to the previous comment: does it mean the charm cannot work with a single unit?

seb4stien (commented):
Left a question on peer_relations.
I feel we should start to have some integration tests with this one to ensure that allow_hosts is working as expected.

haproxy-route-policy-operator/charmcraft.yaml
seb4stien: Do we want it to be open by default? In terms of security we usually go the other way: restrict by default, and open when needed.

Thanhphan1147: Though the point of the charm is also to expose the django admin UI to interact with the rules and requests, making this "localhost" by default will means that the charm won't work when you deploy it until you set the configuration option. 

We can also unset this by default and block the charm until this config is set if we want to take that approach

haproxy-route-policy-operator/src/charm.py
Thanhphan1147: I think this is duplicated with the previous PR, the peer relation should always exists as we declared it in charmcraft.yaml

haproxy-route-policy-operator/src/state/policy.py
Thanhphan1147: I think this is duplicated with the previous PR, the peer relation should always exists as we declared it in charmcraft.yaml

haproxy-route-policy-operator/charmcraft.yaml
javierdelapuente: but why 0.0.0.0? 0.0.0.0 looks like wrong syntax, I mean, it should be localhost, or "*" or just an empty string ""

seb4stien: As discussed in our 1:1: rename allowed-hosts to extra-allowed-hosts that would be empty by default, and give access to the IP coming from the haproxy-route-policy only.

PR #504, opened 2026-05-04
Add pr-best-practices skill

Thanhphan1147:
#### What this PR does

Add a skill that when called will analyze the changes between the first and final iteration of a PR authored by an LLM. The result is a set of good and bad example maintained in `.github/instructions/best-practices.instructions.md`. For example:

```
## Handling Jinja2 templates

Keep rendering logic in charm-state dataclasses or helper builders so templates stay declarative.
State properties that produce ACL conditions should return bare condition strings; the template is responsible for wrapping them in `acl` directive syntax.

<examples>
```

#### Checklist

- [x] I followed the [contributing guide](https://github.com/canonical/is-charms-contributing-guide)
- [x] I added or updated the documentation (if applicable)
- [x] I used AI to assist with preparing this PR
- [x] I added or updated tests as needed (unit and integration)


alithethird (approved):

.github/instructions/best-practices.instructions.md
```diff
@@ -0,0 +1,108 @@
+---
+description: 'Accumulated best practices extracted from LLM-authored PR reviews'
+applyTo: '**'
```
srbouffard, line 3: I wonder if we could have a way to split best practicies into multiple instructions to avoid having a single one that applies too all files.
My worry here is that as we accumulate more and more practices, we will front-load a LOT of context that might not be needed by every agent, every prompt.

Thanhphan1147:
onboarded to team-specific instructions

PR #523, opened 2026-05-23
manage private key in charm instead of lib, add encoder and decoder to load certificates in relation data. update tests

### LINKED ISSUE #366: Share single certificate between units in HA deployment
### Enhancement Proposal

Currently the haproxy-operator charm uses the `TLSCertificateRequiresV4` in `UNIT` mode: https://github.com/canonical/haproxy-operator/blob/120f97f6b352510c218b565ca65e3713921aaa20/haproxy-operator/src/charm.py#L128-L138

This means that in HA deployments, a separate certificate is requested for each unit. Let's Encrypt only permits [a maximum of 5 certificates to be requested for a given domain across 168h](https://letsencrypt.org/docs/rate-limits/#new-certificates-per-exact-set-of-identifiers). This means if there is there is an error with the initial certificate issuing, users must potentially wait several days before requesting a new cert.

We have regularly seen this issue in production use with PS7 ingress environments.

This feature request is to request using `mode.APP` for `TLSCertificateRequiresV4` and sharing the single certificate to other units in the HA deployment.

Thanhphan1147:
Fixes #366 

#### Checklist

- [x] I followed the [contributing guide](https://github.com/canonical/is-charms-contributing-guide)
- [x] I added or updated the documentation (if applicable)
- [x] I used AI to assist with preparing this PR
- [x] I added or updated tests as needed (unit and integration)


haproxy-operator/src/state/tls.py
```diff
@@ -12,18 +14,70 @@
...
+    """A dataclass to represent a certificate and its chain.
+
+    Attributes:
+        shared_certificates: A list of tuples, each containing a certificate and its associated chain.
+    """
+
+    shared_certificates: list[tuple[Certificate, list[Certificate]]]
+
+
+def haproxy_peer_relation_app_data_encoder(value: typing.Any) -> str:
```
alithethird, line 54: imho its better to not do type annotation if we end up putting any in there.

haproxy-operator/src/state/validation.py
```diff
@@ -75,12 +79,14 @@ def wrapper(instance: C, *args: typing.Any) -> None:
                 instance.unit.status = ops.BlockedStatus(str(exc))
                 logger.exception("Not ready to handle TLS.")
                 return None
-            except PrivateKeyNotGeneratedError as exc:
+            except (SharedTLSInformationNotFoundError, PeerRelationNotFoundError) as exc:
```
alithethird, line 82: `PrivateKeyNotGeneratedError` error can still be fired right? Do we let it bubble up and put the charm in error state?

alithethird (approved):

haproxy-operator/src/charm.py
```diff
@@ -416,6 +431,12 @@ def _configure_haproxy_route(
             self._publish_haproxy_route_tcp_proxied_endpoints(
                 haproxy_route_requirers_information, ha_information
             )
+            if peer_relation := self.model.get_relation(HAPROXY_PEER_INTEGRATION):
```
javierdelapuente: Does this only apply to haproxy_route?

haproxy-operator/tests/unit/test_charm.py
```diff
@@ -338,7 +338,8 @@ def test_spoe_auth(monkeypatch: pytest.MonkeyPatch, certificates_integration):
```
javierdelapuente, line 355: Maybe a test with leader False that tests the new use case?

haproxy-operator/src/state/tls.py
```diff
@@ -67,22 +120,55 @@ def from_charm(
...
+                    (certificate.certificate, certificate.chain)
+                    for certificate in provider_certificates
+                ]
+            ),
+            private_key=str(private_key),
+        )
 
+    @classmethod
+    def _from_peer_relation(
+        cls, charm: ops.CharmBase, hostnames: list[str], allow_no_certificates: bool
```
swetha1654, line 147: what is `allow_no_certificates` param for?

Thanhphan1147: it's for when we have only TCP passthrough on the backends, in that case we can't enforce HTTPS and we need to continue the reconcile loop without writing certificates to the units.

haproxy-operator/tests/unit/test_charm.py
Thanhphan1147: added ( by copilot )

haproxy-operator/src/charm.py
Thanhphan1147: added the same for ingress

haproxy-operator/src/state/validation.py
Thanhphan1147: Yes, because the private key is managed by the charm now and we ensure that it exists on every init so not having it here means that something very bad has happened :sweat_smile:

Thanhphan1147: updated the code to make this error more clear

haproxy-operator/src/state/tls.py
Thanhphan1147: Makes sense, but seems like this should be a discussion on the overall code style of the repo, let's continue internally and I'll keep the type hints aligned with the rest of the code for now

haproxy-operator/src/state/tls.py
Thanhphan1147: I'll resolve this so that the PR can be automerged, let me know if we should continue this discussion

PR #538, opened 2026-06-05
feat: add port-range attribute to haproxy-route-tcp interface

### LINKED ISSUE #525: Add a `port-range` attribute to `haproxy-route-tcp` to allow mapping from mutliple frontend ports to multiple backend ports
### Enhancement Proposal

* We need to be able to allow the requirer to specify a range of port that it wants to be mapped 1-to-1 from the frontend. For example: if a range of `10500-10600` is requested, then every port in that range on the frontend is mapped to the same port on the backend, e.g `<haproxy_ip>:10512` will map to `<backend_server_ip>:10512`
* Conflicts will need to be checked for the entire port range, if one port in the range is conflicting then the entire backend will be considered invalid.
* The `port` part of each server entries in the backend will be removed if `port-range` is set.
* every port in the range will need to be used in `open_ports` call at the end of the reconcile loop.

### Impact

High (The feature has short-term technical value)

### Impact Rationale

_No response_

seb4stien:
## Description

Implements #525 — adds a `port_range` attribute to the `haproxy-route-tcp` interface, allowing requirers to expose a range of ports (e.g. `"10500-10600"`) that are mapped 1-to-1 from frontend to backend.

### How it works

When a frontend uses `bind [::]:10500-10600`, HAProxy passes the destination port of each incoming connection to the backend. Backend server entries with **no port specified** inherit the frontend destination port, achieving 1-to-1 port mapping without explicit per-port configuration.

### Changes

- **Library** (`v1`, LIBPATCH→5): `parse_port_range()` helper, `port` made optional, `port_range: Optional[str]` field added with XOR validator (exactly one of `port`/`port_range` must be set), range-aware conflict detection expands ranges to individual ports before checking overlaps.
- **State** (`haproxy_route_tcp.py`): `HaproxyRouteTcpServer.port` is now `Optional[int]`; `HAProxyRouteTcpFrontend` gains `port_range: Optional[tuple[int,int]]` field and `all_ports` property returning all covered ports.
- **State** (`haproxy_route.py`): frontend grouping key is now a string (port number or range string); conflict detection and valid-frontend filtering use `all_ports` to expand ranges.
- **Template**: server macro omits `:port` when port is `None`; frontend section name uses `start_end` and bind line uses `start-end` (HAProxy syntax).
- **Charm**: `set_ports` expands port ranges; proxied endpoints for range frontends are published as `"host:start-end"`.
- **Tests**: 12 new library-level tests, 2 new proxied-endpoint tests, updated `build_haproxy_route_tcp_relation` fixture to accept `port_range`.

### Testing

```
254 tests pass, lint clean
```

Closes #525

### DIFF
--- a/docs/release-notes/artifacts/pr0538.yaml
+++ b/docs/release-notes/artifacts/pr0538.yaml
@@ -1,0 +1,14 @@
+# --- Release notes artifact ----
+
+schema_version: 1
+changes:
+  - title: Added port_range support to haproxy-route-tcp interface
+    author: github-actions[bot]
+    type: feature
+    description: The haproxy-route-tcp interface now supports a port_range attribute (e.g. "10500-10600") for 1-to-1 port mapping using HAProxy's bind range syntax.
+    urls:
+      pr: https://github.com/canonical/haproxy-operator/pull/538
+      related_doc:
+      related_issue: https://github.com/canonical/haproxy-operator/issues/525
+    visibility: public
+    highlight: false
--- a/haproxy-operator/lib/charms/haproxy/v1/haproxy_route_tcp.py
+++ b/haproxy-operator/lib/charms/haproxy/v1/haproxy_route_tcp.py
@@ -708,10 +708,10 @@ class TcpRequirerApplicationData(_DatabagModel):
 
     @model_validator(mode="after")
     def validate_port_xor_port_range(self) -> "Self":
-        """Ensure exactly one of port or port_range is set.
+        """Ensure port and port_range are not both set.
 
         Raises:
-            ValueError: When neither or both port and port_range are set,
+            ValueError: When both port and port_range are set simultaneously,
                 or when port_range has an invalid format.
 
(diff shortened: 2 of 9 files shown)


seb4stien:
The `test_haproxy_route_tcp.py` integration test failed with a `TimeoutError: wait timed out after 600s`. Looking at the logs, this was caused by infrastructure instability on the self-hosted runner — the Juju controller API was returning `connection refused` on `127.0.0.1:17070`, preventing the model from becoming active:

```
machine-0: 10:29:29 ERROR juju.worker.dependency "api-caller" manifold worker returned unexpected error: [6c58f1] "machine-0" cannot open api: unable to connect to API: dial tcp 127.0.0.1:17070: connect: connection refused
```

This is unrelated to the changes in this PR (adding `port_range` support). The test runs fine on `main` and the failure is a known pattern with the Canonical self-hosted runners. All other integration tests passed.

PR #552, opened 2026-06-12
Revert snap to amd64 only and update charm-ci

javierdelapuente:
#### What this PR does

Revert snap to be only amd64. The charm is not adapted, and the publish fails, so no need for now.

Update charm-ci to fix the publish.

#### Checklist

- [ ] I followed the [contributing guide](https://github.com/canonical/is-charms-contributing-guide)
- [ ] I added or updated the documentation (if applicable)
- [ ] I updated `docs/changelog.md` with user-relevant changes
- [ ] I added a [change artifact](../docs/release-notes/template/_change-artifact-template.yaml) for user-relevant changes in `docs/release-notes/artifacts`. If no change artifact is necessary, I tagged the PR with the label `no-release-note`.
- [ ] I used AI to assist with preparing this PR
- [ ] I added or updated tests as needed (unit and integration)

### DIFF
--- a/haproxy-spoe-auth-snap/snapcraft.yaml
+++ b/haproxy-spoe-auth-snap/snapcraft.yaml
@@ -12,7 +12,8 @@ description: |
 
 platforms:
   amd64:
-  arm64:
+    build-on: [amd64]
+    build-for: [amd64]
 
 grade: stable
 confinement: strict
--- a/haproxy-spoe-auth-snap/snapcraft.yaml
+++ b/haproxy-spoe-auth-snap/snapcraft.yaml
@@ -10,11 +10,6 @@ description: |
   A SPOE (Stream Processing Offload Engine) agent for HAProxy that provides
   authentication capabilities.
 
-platforms:
-  amd64:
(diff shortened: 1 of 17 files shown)


haproxy-spoe-auth-snap/snapcraft.yaml
```diff
@@ -12,7 +12,8 @@ description: |
 
 platforms:
   amd64:
-  arm64:
+    build-on: [amd64]
+    build-for: [amd64]
```
swetha1654: do we need to explicitly specify this? Removing arm64 is not sufficient?

swetha1654 (approved):

haproxy-spoe-auth-snap/snapcraft.yaml
javierdelapuente: Thanks for the heads up. I am going to put the two snapcraft.yaml exactly as what was before (before the charm-ci) so the publish for the snap should work and we can then iterate on it (I am not going to put arm64 in any of the snaps).

PR #609, opened 2026-07-17
Feat/get configuration action

minulo:
#### What this PR does
Add a get-configuration Juju action so operators can inspect the effective haproxy configuration for debugging without shell access to the unit. 

- source=disk (default): return the applied /etc/haproxy/haproxy.cfg
- source=relations: preview the configuration the next reconcile would generate from the current haproxy-route relation data (read-only; no file writes or service reload)
- Log a warning when the effective configuration matches the default (no proxy backends configured) and, in relations mode, when a haproxy-route-policy relation makes the policy backend converge asynchronously (source=disk may also face the async issue if run before the haproxy-route-policy get written to file)
- Refactored shared template-context and render-to-string helpers so the reconcile and preview paths cannot drift

#### Why we need it
The haproxy charm creates configuration based on its haproxy-route, haproxy-route-policy, and haproxy-route-tcp relations, but there is no way to inspect the effective configuration without shell access to the unit. This spec proposes a get-configuration Juju Action that returns the charm’s configuration out to stdout by both reading from disk and calculating the potential resulting configuration based on current relations. 

#### Checklist

- [ x] I followed the [contributing guide](https://github.com/canonical/is-charms-contributing-guide)
- [ ] I added or updated the documentation (if applicable)
- [ x] I used AI to assist with preparing this PR
- [ x] I added or updated tests as needed (unit and integration)


haproxy-operator/src/charm.py
```diff
@@ -284,10 +400,13 @@ def _on_http_backend_removed(self, _: HTTPBackendRemovedEvent) -> None:
...
-        """Render the haproxy config and restart the service."""
-        self.unit.status = ops.MaintenanceStatus("Configuring haproxy.")
-        charm_state = CharmState.from_charm(
+    def _charm_state(self) -> CharmState:
+        """Build the charm state from the current charm and its providers.
+
+        Returns:
+            The charm state component.
+        """
+        return CharmState.from_charm(
```
Thanhphan1147: why is this needed? Is it because the method passed the recommended complexity?

If we're moving the initialization of the charm state here, then we should probably also do error handling here instead of letting it flow up to _reconcile and to the decorator, which is a bigger refactoring question.

So I think we can just ignore the complexity rule for now

haproxy-operator/src/charm.py
```diff
@@ -717,6 +841,159 @@ def _on_get_proxied_endpoints_action(self, event: ActionEvent) -> None:
...
+                CharmStateValidationBaseError,
+                HaproxyRouteIntegrationDataValidationError,
+            ) as exc:
+                event.fail(f"Failed to recompute configuration from relations: {exc}")
+                return
+            if self.haproxy_route_policy.relation is not None:
+                event.log(
+                    "A haproxy-route-policy relation is present; the policy backend in this "
+                    "preview reflects the policy charm's current output and converges "
+                    "asynchronously. Use source=disk for the authoritative applied configuration."
```
Thanhphan1147: Not sure what it means here by "converging asynchronously". 

If the `haproxy-route-policy` relation is present, then the relation data of that relation should be the source of truth about which will get written to disk, simple as that.

haproxy-operator/src/charm.py
```diff
@@ -717,6 +841,159 @@ def _on_get_proxied_endpoints_action(self, event: ActionEvent) -> None:
 
         event.set_results({"endpoints": json.dumps(proxied_endpoints)})
 
+    def _on_get_configuration_action(self, event: ActionEvent) -> None:
```
Thanhphan1147, line 721: I think we're trying to do too much here, which introduces a lot of complexity.

From what I understand, we try to converge the output of both `source=relation` and `source=disk` into a big blob of text, then we do some filtering on that blob of text, which sounds very fragile.

Personally I think we should never parse plain text, so `source=disk` should not have any filtering at all.

When we're getting it from the relation data, since we're dealing with structured data, then parsing might make sense.

Though thinking more about what we're trying to do, I'm having doubts as to whether showing the config per-backend is really necessary. Maybe what we can do is start with `source=disk` only to keep things simple, then evaluate `source=relation` in a separate PR.

haproxy-operator/src/charm.py
```diff
@@ -108,6 +108,121 @@ def _validate_port(port: int) -> bool:
     return 0 <= port <= 65535
 
 
+def _strip_shared_boundaries(configuration: str, reference: str) -> str:
```
Thanhphan1147: We should know how many lines to trim out since that part is always the same right? I think we don't need a dedicated method for this and just strip out the first X lines. Unless I'm mistaken and that's actually not easily doable

minulo: I knew that the part (i.e: global) is always the same, however, I don't know if the "always-the-same" template would ever change or not, so I make the function drift-resistant by explicitly referring to the default templates as the source of truth to know what to get rid of. 

Thus, if the base template ever changes (a new global line, a renamed section, etc.), the default render changes with it and the stripping stays correct automatically

haproxy-operator/src/charm.py
minulo: I didn't refactored because of complexity rule (I just found out about it from your comment and saw the linter complexity reccomendation). 

The reason why I refactored this out is twofold:
- Firstly, I prefer Python functions without side effects. 
- Secondly, I want to prevent any kind of drift if possible, and as I find myself needing to call CharmState.from_charm with its list of arguments in different places, I figured I might as well refactored into a new function for reusability and to prevent potential drifting (of arguments fed to CharmState.from_charm for example).  

As for your error handling point, I figured that each places that I call _charm_state may have different errors they cared about, so I didn't want to have error handling in the function. I am not sure if this is best practices or not, but at least it makes sense from my perspective.

haproxy-operator/src/charm.py
minulo: The code right now is not converging the output of both `source` mode, but rather one or another. Also, your understanding is correct in terms of backend filtering. The filtering is based on the start of the blob of text is matching the backend and then the blob of text ended if there is a new line symbol. While I think it is drift resistant, I am not sure whether it is impossible for a backend description to contain a new line element or not. 

With that being said, I think even at 20 relations, it is still very readable even without the backend filtering function.

I am willing to open another PR for a separate `source=relation` feature, but this time, it would definitely simplify things if we only keep `source=disk`. Also, it helps with the converge asynchrously issue below.

haproxy-operator/src/charm.py
minulo: When self.haproxy_route_policy.relation is present, the policy backend is computed as an output of a separate Postgres-backed charm. Thus, `source=relations` can previews stale/not-yet-settled values. Given the use case of this action to use for debugging (I can see someone who would add relation and then immediately run the action which can show stale data), thus, I decided to output this log so the user of this use case won't be confused. 

As you suggested above, getting rid of `source=relation` would greatly simplify this. There may still be stale data if the user use the action fast enough, but I doubt it.

minulo: That is what I meant by converge asynchronously.

haproxy-operator/src/charm.py
Thanhphan1147: Makes sense, let's see the PR with only `source=disk` first and leave the filtering for later

haproxy-operator/src/charm.py
Thanhphan1147: It's a bit specific to the policy charm but the data from the `haproxy-route-policy` relation is not "stale" but actually the opposite. When the  `haproxy-route-policy` relation is present, only the data coming from that relation is taken into account and written to disk. ( on the policy charm side there's an approval process to decide what gets written to disk on the haproxy charm ).

But let's simplify by not including `source=relation` for now and deal with this later.

haproxy-operator/src/charm.py
Thanhphan1147: Seems like we need to discuss about this more, since I feel like adding logic for this is too much, but we can't reliably maintain a constant with an implicit constraint as well ( maybe we can even read the template file and determinte how far we need to trim, ... )

But let's skip this for now since it's not a very important thing to have.

haproxy-operator/src/charm.py
Thanhphan1147: Though you are basically wrapping a method with another method without adding anything else besides abstracting away the parameters ( essentially renaming it ), which I'd argue to be an anti-pattern since now we have to look one step further to  understand how state is initialized, what parameters does that state needs, etc. It's reducing maintainability imo.

I'd understand if we bundle error handling into this method that way the wrapper add some value other than just abstracting away the parameters, and I'd argue that adding error handling here make sense because the overall design of the charm is hollistic, so there should be no reason ( aside from in action handlers ) that different parts of the code cares about different errors. What if the error that it didn't care about get raised?

minulo: I have reverted the changes. After getting rid of source=relations, _charm_state doesn't really need that refactoring.

haproxy-operator/tests/integration/test_actions.py
```diff
@@ -5,11 +5,36 @@
 """Integration tests for haproxy charm actions."""
 
 import json
+import time
 
 import jubilant
 import pytest
 
 
+def _integrate_haproxy_route(juju: jubilant.Juju, provider_endpoint: str, requirer: str) -> None:
```
Thanhphan1147: I don't think this is needed.

To go a bit into details on how integration tests are structured, test modules are executed in parallel by `charm-ci` ( defined in `.github/workflows` ), so ideally each module should have only one test method to avoid having to handle teardown of each test method, ensuring a clean state after each test, etc. 

We organized the tests this way purely for simplicity and being able to run them in parallel ( not having to deal with teardown and other stuffs that I mentioned above )

So instead of adding another test method I'd refactor the `test_get_proxied_endpoints_action` into a generic `test_action` method and run both actions there. And remove this helper method

haproxy-operator/tests/unit/test_charm.py
```diff
@@ -470,3 +471,83 @@ def test_spoe_auth_invalid_data(monkeypatch: pytest.MonkeyPatch, certificates_in
     assert render_file_mock.call_count == 0
     assert out.unit_status.name == ops.testing.BlockedStatus.name
     assert spoe_auth_relation.remote_app_name in out.unit_status.message
+
+
+@pytest.mark.usefixtures("systemd_mock", "mocks_external_calls")
+class TestGetConfigurationAction:
```
Thanhphan1147: Can I ask you to keep the format of the unit tests aligned with other tests in the repo? ( isolated method, with "arrange", "act" and "assert" and any fixtures used is scoped to only that test method )

haproxy-operator/tests/unit/test_charm.py
```diff
@@ -470,3 +471,83 @@ def test_spoe_auth_invalid_data(monkeypatch: pytest.MonkeyPatch, certificates_in
...
+    """Test "get-configuration" Action."""
+
+    def test_returns_configuration(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        """
+        arrange: mock the config file on disk with known content.
+        act: trigger the get-configuration action.
+        assert: the on-disk file content is returned unchanged.
+        """
+        content = "global\n    maxconn 4096\n\nfrontend default\n    bind :80\n"
+        monkeypatch.setattr(charm_module, "file_exists", lambda _: True)
```
Thanhphan1147: can't you do `monkeypatch.setattr("charm.file_exists", MagicMock(return_value=True)`?

Thanhphan1147: That way you won't need the `import charm as charm_module`

haproxy-operator/tests/unit/test_charm.py
```diff
@@ -470,3 +471,83 @@ def test_spoe_auth_invalid_data(monkeypatch: pytest.MonkeyPatch, certificates_in
...
+            charm_module.HAProxyService,
+            "render_default_config",
+            lambda self, _: "global\n    maxconn 4096\n",
+        )
+        context = ops.testing.Context(HAProxyCharm)
+        state = ops.testing.State(leader=True)
+
+        context.run(context.on.action("get-configuration"), state)
+
+        assert not any("default configuration" in log.lower() for log in context.action_logs)
```
Thanhphan1147: Can you group structurally similar tests with a `@pytest.mark.parametrize`?

haproxy-operator/charmcraft.yaml
```diff
@@ -146,6 +146,11 @@ actions:
         description: |
           The name of the backend for which to get the endpoints.
           If no backend with this name is present, an empty list is returned.
+  get-configuration:
+    description: |
+      Return the rendered haproxy configuration currently on disk
+      (/etc/haproxy/haproxy.cfg) that haproxy is running. Intended for
+      debugging purposes.
```
Thanhphan1147:
```suggestion
      Return the rendered haproxy configuration currently on the haproxy unit. Intended for debugging purposes.
```

haproxy-operator/tests/integration/test_actions.py
minulo: Thanks for telling me. I will look into this...

haproxy-operator/src/charm.py
```diff
@@ -717,6 +718,58 @@ def _on_get_proxied_endpoints_action(self, event: ActionEvent) -> None:
...
+                    self._ingress_provider,
+                    self._ingress_per_unit_provider,
+                    self.haproxy_route_provider,
+                    self.haproxy_route_tcp_provider,
+                    self.reverseproxy_requirer,
+                    self.haproxy_route_policy,
+                )
+            )
+        except CharmStateValidationBaseError:
+            return False
```
Thanhphan1147: What does catching that error here means? can you add a small comment and add a `logger.warning` line to explain why we consider that haproxy is not running the default configuration if we have an issue parsing the charm_state?

haproxy-operator/tests/unit/test_charm.py
minulo: Yes, I will do so.

haproxy-operator/src/charm.py
```diff
@@ -717,6 +718,58 @@ def _on_get_proxied_endpoints_action(self, event: ActionEvent) -> None:
...
+        disk or reload the service.
+
+        Args:
+            event: Juju event
+        """
+        if not file_exists(HAPROXY_CONFIG):
+            event.fail(
+                f"HAProxy configuration file {HAPROXY_CONFIG} does not exist yet. "
+                "Ensure the charm is configured and integrated before running this action."
+            )
```
Thanhphan1147:
```suggestion
            event.fail(
                f"HAProxy configuration file at {HAPROXY_CONFIG} not found. "
            )
```

The config file should always exist, haproxy won't start without it

haproxy-operator/tests/unit/test_charm.py
minulo: Definitely.

haproxy-operator/src/charm.py
```diff
@@ -717,6 +718,58 @@ def _on_get_proxied_endpoints_action(self, event: ActionEvent) -> None:
...
+            return
+        configuration = read_file(HAPROXY_CONFIG)
+
+        if self._configuration_is_default(configuration):
+            event.log(
+                "The HAProxy configuration matches the default configuration. This usually "
+                "means no proxy backends are configured (e.g. no haproxy-route, ingress, or "
+                "reverseproxy relations)."
```
Thanhphan1147:
```suggestion
                "means no proxy backends are configured."
```

haproxy-operator/src/charm.py
minulo: Yeah, now that I think about it, maybe using try catch for a comparison purpose is definitely not best practice and it hides information away as well if the parsing fails... 

I think it will make the best sense to move error handling up to _on_get_configuration_action() as this is an action handler so it make sense to handle error there. I believe this lines up with what we discussed before and I think by doing so it preserves the most amount of information in case charm_state fails to parse. 

I will include this in the next few PRs so that you can see what I mean.

haproxy-operator/src/charm.py
```diff
@@ -717,6 +718,63 @@ def _on_get_proxied_endpoints_action(self, event: ActionEvent) -> None:
...
+            event.fail(f"HAProxy configuration file at {HAPROXY_CONFIG} not found. ")
+            return
+        configuration = read_file(HAPROXY_CONFIG)
+
+        try:
+            configuration_is_default = self._configuration_is_default(configuration)
+        except CharmStateValidationBaseError:
+            event.log(
+                "Could not determine whether this is the default configuration because the "
+                "charm state is invalid; see `juju status` for the blocking condition."
```
Thanhphan1147:
```suggestion
                "charm state is invalid."
```

haproxy-operator/tests/integration/test_actions.py
```diff
@@ -57,28 +58,32 @@ def test_get_proxied_endpoints_action(
         "https://ok3.haproxy.internal/v2",
     }
 
-    # Test without backend param
+    # Test without a backend param (filter)
```
Thanhphan1147, line 61:
```suggestion

```

minulo:
> Looks good, I think there's only the refactoring and grouping of the tests left until we're good to merge

Hey @Thanhphan1147, I believe I have already refactored to test_action that combines both Juju action testing and I used parametrize to group the two existing default-config-warning tests into test_default_configuration_warning. I'm not sure what your comment is point towards. Are you perhaps referring to combining the TestsGetProxiedEndpointsAction with my TestGetConfigurationAction class and refactor to be one class that tests all action?

Thanhphan1147:
@minulo if you look at the unit tests in this repo we don't put them under a test class, since the module should provide sufficient logical grouping already. Can you remove the test class and just keep the tests as isolated methods?

minulo:
> @minulo if you look at the unit tests in this repo we don't put them under a test class, since the module should provide sufficient logical grouping already. Can you remove the test class and just keep the tests as isolated methods?

@Thanhphan1147 I see... I will then turn it to individual functions then. I assumed using a class would have been the correct choice given the tests for the other action, TestGetProxiedEndpointsAction, were constructed as a class when the other unit tests were all functional. I assume that there are some reason for this that I am not aware of.

Thanhphan1147:
@minulo oh I think the other tests in that file just happen to not follow the convention of the repo for some reason :smile: We'll have to fix that too.

Thanhphan1147 (approved):
Good to merge once the tests are refactored

minulo:
> @minulo oh I think the other tests in that file just happen to not follow the convention of the repo for some reason 😄 We'll have to fix that too.

Cool, I can handle it in different PR, probably next sprint if you want.

